PR #29, opened 2023-09-04
Use translation keys for entity names

dknowles2:


dknowles2:
This doesn't seem to work correctly for custom integrations.

PR #41, opened 2024-01-16
Update README.md

savdagod:
Added section to help users achieve better signal by switching to an external antenna that is mounted on the outside of the hopper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,9 @@ Power on your grill and it should be discovered automatically over Bluetooth. On
 initiate the setup process, it will ask for your exact grill model so we can properly
 communicate with it. No cloud access necessary!
 
+## Tips to get a better connection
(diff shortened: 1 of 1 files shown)


dknowles2:
This sounds like it's specific to a certain grill model. There's definitely no access to an antenna port through my grill's hopper, for example. (I just had the whole thing pulled apart a few weeks ago since my cover blew off during a rain storm and my auger was completely clogged with wet pellets)

Maybe rephrase this containing a specific grill model, or make it generic enough so that it reads as a possibility?

savdagod:
Hmm maybe it’s specific to controllers. I purchased the trapezoid controller and replaced the stock on that came with my grill. The ESP chip itself has the antenna port. Looks like this https://www.digikey.com/en/products/detail/espressif-systems/ESP32-WROOM-32UE-N4/11613176?utm_adgroup=&utm_source=google&utm_medium=cpc&utm_campaign=PMax%20Shopping_Product_Medium%20ROAS%20Categories&utm_term=&utm_content=&utm_id=go_cmp-20223376311_adg-_ad-__dev-m_ext-_prd-11613176_sig-CjwKCAiAkp6tBhB5EiwANTCx1PwNaDVtK1JoO_NN0R2WUS_p6htdcbR_VImBXx3WQFI_IXJp101zUhoCdWUQAvD_BwE&gad_source=1.

im thinking maybe only the newer or trapezoid controllers use a chip with a removable antenna. I will reword it to mention that.

dknowles2:
So these tips are specifically for a modded grill then? That's worth mentioning. I'm not sure there exists an aftermarket controller module for my smoker, which is why I'm hesitant to include this.

savdagod:
It’s not modded in terms of a homebrew hack. PitBoss sells the WiFi/Bluetooth controller module that you just drop into your existing grill. It’s the same modules that would come on the grill if you bought it with the features “built-in”.

PR #118, opened 2025-09-04
Recipe Support

mccordscvs:
- **Added a sensor for recipe step and recipe time.** 
- Do not install the primer switch if the "turn-primer-motor-on" command is not available. It's only available on 110/137 grills.
- Do not install probe 1 targets if prob target commands are not available. 36/137
- Do not install the recipe sensors if recipe option is not supported. 6/137
- Various Refactoring


dknowles2:
Please split this into separate PRs that address each item individually.

PR #121, opened 2025-09-07
Add Recipe Sensors

mccordscvs:
Added a recipeStep sensor and a recipeTime sensor for grills that support it.

### DIFF
--- a/custom_components/pitboss/sensor.py
+++ b/custom_components/pitboss/sensor.py
@@ -62,7 +62,7 @@ class RecipeSensorEntityDescription(SensorEntityDescription):
 RECIPE_ENTITY_DESCRIPTIONS = (
     RecipeSensorEntityDescription(
         key="recipeTime",
-        name="Recipe Name",
+        name="Recipe Time",
         icon="mdi:clock-outline",
         device_class=SensorDeviceClass.DURATION,
         native_unit_of_measurement=UnitOfTime.SECONDS,
@@ -110,15 +110,6 @@ class BaseSensorEntity(BaseEntity, SensorEntity):
     ) -> None:
         super().__init__(coordinator, entry_unique_id)
         self.entity_description = entity_description
-        if isinstance(entity_description, ProbeSensorEntityDescription):
-            self.probe_number: int = entity_description.probe_number
-            self._attr_unique_id = f"probe{self.probe_number}_{entry_unique_id}"
-        else:
-            self._attr_unique_id = f"{entity_description.key}_{entry_unique_id}"
-
-    @property
-    def available(self) -> bool:
-        return bool(self.coordinator.data) and super().available
 
     @property
     def native_value(self) -> int | None:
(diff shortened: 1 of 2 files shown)


custom_components/pitboss/sensor.py
```diff
@@ -51,32 +51,87 @@ class ProbeSensorEntityDescription(SensorEntityDescription):
...
+    """Describes a PitBoss recipe sensor."""
+
+    key: Literal["recipeTime", "recipeStep"]
+    state_class: SensorStateClass = SensorStateClass.MEASUREMENT
+
+
+RECIPE_ENTITY_DESCRIPTIONS = (
+    RecipeSensorEntityDescription(
+        key="recipeTime",
+        name="Recipe Name",
```
dknowles2, line 65: Should this be "Recipe Time"?

custom_components/pitboss/sensor.py
```diff
@@ -51,32 +51,87 @@ class ProbeSensorEntityDescription(SensorEntityDescription):
...
+        self.entity_description = entity_description
+        if isinstance(entity_description, ProbeSensorEntityDescription):
+            self.probe_number: int = entity_description.probe_number
+            self._attr_unique_id = f"probe{self.probe_number}_{entry_unique_id}"
+        else:
+            self._attr_unique_id = f"{entity_description.key}_{entry_unique_id}"
+
+    @property
+    def available(self) -> bool:
+        return bool(self.coordinator.data) and super().available
```
dknowles2, line 121: Should this just go in `BaseEntity`?

custom_components/pitboss/sensor.py
```diff
@@ -51,32 +51,87 @@ class ProbeSensorEntityDescription(SensorEntityDescription):
...
-        entity_description: ProbeSensorEntityDescription,
+        entity_description: PBSensorEntityDescription,
     ) -> None:
         super().__init__(coordinator, entry_unique_id)
-        self.entity_description: ProbeSensorEntityDescription = entity_description
-        self._attr_unique_id = (
-            f"probe{entity_description.probe_number}_{entry_unique_id}"
-        )
+        self.entity_description = entity_description
+        if isinstance(entity_description, ProbeSensorEntityDescription):
```
dknowles2, line 113: Instead of the `isinstance` check, you can just move these into the initializer for each of the sensor entity classes.

custom_components/pitboss/sensor.py
mccordscvs: Yes, I was in the shower and thought I didn't change the name. Autocode did "recipeName" for the key instead of "recipeStep." I changed the key but forgot the name. I'm glad you haven't pulled it yet. I will fix it when I get back home.

custom_components/pitboss/sensor.py
mccordscvs: Yeah, that's probably better. I fell into the trap of trying to cut lines of code weaher it's the best idea.

PR #122, opened 2025-09-07
Disable target probe1

mccordscvs:
Only a few control boards support 'set-probe-1-temperature'. Only init the entity if it's available.


custom_components/pitboss/number.py
```diff
@@ -92,7 +103,7 @@ def available(self) -> bool:
                 data.get(self.entity_description.matching_probe_key) is not None
                 and super().available
             )
-        return super().available
+        return super().available and bool(data)
```
dknowles2, line 106: I don't think we need the `bool(data)` check here?

custom_components/pitboss/number.py
```diff
@@ -46,14 +52,19 @@ async def async_setup_entry(
     """Setup number platformm."""
     coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
     assert entry.unique_id is not None
-    entities = [
-        TargetProbeTemperature(coordinator, entry.unique_id, PROBE_1_DESCRIPTION)
-    ]
-    if "set-probe-2-temperature" in coordinator.api.spec.control_board.commands:
+    entities: list[TargetProbeTemperature] = []
+    if "set-probe-1-temperature" in (
+        api := coordinator.api.spec.control_board.commands
```
dknowles2, line 57: Just assign this on the line before. I'd recommend calling it `available_commands` instead.

mccordscvs: Yes, thats a more descriptive name.

mccordscvs:
Is there any more problems with this that I can address?

PR #258, opened 2026-06-06
Map Louisiana Black Label (LBL) prefix to PBL control board

entroversal:
## Summary

Louisiana Grills is a Dansons sibling brand of Pit Boss that shares the same control boards but advertises over BLE with brand-specific name prefixes. The manifest already discovers `LBL-*` devices, but the config flow derives the control board from the name prefix, yielding `"LBL"` — a board `pytboss` doesn't know. `grills.get_grills(control_board="LBL")` returns an empty list, so the flow aborts with `unknown_grill` right after discovery.

In other words: the grill is found, but cannot be added.

## Change

Refactor the existing `PBL2` -> `PBL3` special case into a small `CONTROL_BOARD_ALIASES` table and add `LBL` -> `PBL`. The Black Label series uses the WiFi+BT `PBL` board on the Pit Boss side.

```python
CONTROL_BOARD_ALIASES = {
    "PBL2": "PBL3",
    "LBL": "PBL",  # Louisiana Black Label series (verified on an LG800BL)
}
...
control_board = self._device_id.split("-")[0]
control_board = CONTROL_BOARD_ALIASES.get(control_board, control_board)
```

## Verification

Tested end to end on a **Louisiana Grills 800 Black Label (LG800BL)**:

- Confirmed independently with `pytboss` over BLE that the grill connects with model `PBV4PS2` (a `PBL`-board model) — read firmware `0.2.3` and a live state stream (probe/grill/smoker temps, error flags).
- With this patch, the config flow now lists the `PBL` models instead of aborting, and the device adds and connects over Bluetooth (local) in Home Assistant.

Because the Black Label line (800/1000/1200) shares the one control board, this single mapping covers the whole series.

## Not covered

Other Louisiana prefixes the manifest matches (`LFS`) and the Pit Boss `PBX1` prefix also resolve to unknown boards and still abort. I didn't map them since I can't verify them against hardware — happy to add them in a follow-up if someone can confirm the correct boards.

`ruff check` and `ruff format --check` pass.

### DIFF
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -13,6 +13,15 @@ from pytboss import grills
 
 from .const import ALL_PROTOCOLS, DEFAULT_PROTOCOL, DOMAIN, LOGGER
 
+# Some BLE-name prefixes don't directly match a pytboss control board. Pit Boss
+# and its sibling Dansons brand Louisiana Grills sometimes advertise under a
+# prefix that differs from the control board name pytboss expects. Map those
+# prefixes to the equivalent control board so discovery can resolve a model.
+CONTROL_BOARD_ALIASES = {
+    "PBL2": "PBL3",
+    "LBL": "PBL",  # Louisiana Black Label series (verified on an LG800BL)
+}
+
 
 class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
     """Config flow for PitBoss."""
@@ -86,8 +95,7 @@ class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
     ) -> ConfigFlowResult:
         """Show the more_info form."""
         control_board = self._device_id.split("-")[0]
-        if control_board == "PBL2":
-            control_board = "PBL3"
+        control_board = CONTROL_BOARD_ALIASES.get(control_board, control_board)
         models = [g.name for g in grills.get_grills(control_board=control_board)]
         if not models:
             return self.async_abort(


entroversal:
Heads up: I noticed the native LG0800BL spec exists but is in UNSUPPORTED_MODELS ('bad parsing routine'). This PR is a working interim mapping; happy to instead help fix the native LBL parser in pytboss if you'd prefer that route.

dknowles2:
> Heads up: I noticed the native LG0800BL spec exists but is in UNSUPPORTED_MODELS ('bad parsing routine'). This PR is a working interim mapping; happy to instead help fix the native LBL parser in pytboss if you'd prefer that route.

Yes, the existing test fixtures weren't working with the parsing routines. Since I don't have that grill model I don't have a way of fixing it. If you want to take a stab at it I'd appreciate it. Otherwise some example response data would help me be able to fix the tests.

dknowles2 (commented):
### Code Review Summary

**Verdict: Looks Good**

- **Correctness & Functionality:**  already matches  devices for BLE discovery, but the config flow previously failed when querying  for an  control board. Mapping  ->  cleanly resolves this and allows users with Louisiana Black Label grills to select their model and complete setup.
- **Code Quality:** Refactoring the  ->  special case into  makes the mapping extensible for future board aliases.
- **Verification:** All CI checks (Hassfest, HACS validation, build, mypy, ruff) and local tests pass. Hardware verification on an LG800BL was confirmed by author.

Looks good to merge! @dknowles2 has final say.

dknowles2 (commented):
### Code Review Summary

**Verdict: Looks Good**

- **Correctness & Functionality:** `manifest.json` already matches `LBL-*` devices for BLE discovery, but the config flow previously failed when querying `pytboss` for an `LBL` control board. Mapping `LBL` -> `PBL` cleanly resolves this and allows users with Louisiana Black Label grills to select their model and complete setup.
- **Code Quality:** Refactoring the `PBL2` -> `PBL3` special case into `CONTROL_BOARD_ALIASES` makes the mapping extensible for future board aliases.
- **Verification:** All CI checks (Hassfest, HACS validation, build, mypy, ruff) and local tests pass. Hardware verification on an LG800BL was confirmed by the author.

Looks good to merge! @dknowles2 has final say.

trailro:
Some data that might help unblock this. I don't have an LBL grill, but I run this integration daily on a **PB1600PS1 (PBL board, 4 probes)** and looked into why the native LBL models are flagged as having a bad parsing routine.

### The LBL `status_function` is internally inconsistent

Comparing `LG0800BL` (board `LBL`) with `PB1600PS1` (board `PBL`) in `grills.json`. LBL parses temperatures inline (it has no `temperatures_function`) and then reads status flags starting at index 21:

```js
p1Target:      convertTemperature(parts, 2),
p1Temp:        convertTemperature(parts, 5),
p2Temp:        convertTemperature(parts, 8),
p3Temp:        convertTemperature(parts, 11),
p4Temp:        convertTemperature(parts, 14),
smokerActTemp: convertTemperature(parts, 14),   // same index as p4Temp
moduleIsOn: parts[21] === 1,
err1:       parts[22] === 1,
err2:       parts[23] === 1,
...
switch (parts[23]) {           // also used as err2
  case 1: status.grillSetTemp = convertTemperature(parts, 20); break;
  case 2: status.grillTemp    = convertTemperature(parts, 20); break;
}
```

Two concrete problems:

1. `smokerActTemp` reads index 14, the same index as `p4Temp` (PBL uses 17).
2. The grill temperature is read at index 20, which spans bytes 20-22, but `moduleIsOn` is read at 21 and `err1` at 22. Those overlap, so they cannot both be right.

### The indices look like PBL's, shifted down by 3

| field | LBL | PBL |
|---|---|---|
| moduleIsOn | 21 | 24 |
| err1 | 22 | 25 |
| primeState | 35 | 38 |
| recipeStep | 37 | 40 |
| recipeTime | 38/39/40 | 41/42/43 |

The offset is exactly 3, i.e. one temperature field. That is consistent with the LBL routine being the PBL layout with one temperature dropped (which is also why `smokerActTemp` ended up colliding with `p4Temp`) and everything after it shifted down.

### Validated against a real PBL frame

Raw `FE0B` frame captured over BLE from my PB1600PS1 shortly after lighting it (igniter, fan and auger all running):

```
FE0B000100090600090600090600090600090600000204020100000000000000000001010100000200000000FF
```

45 bytes. Decoded with the PBL indices:

```
moduleIsOn  parts[24] = 1
fanState    parts[34] = 1
hotState    parts[35] = 1
motorState  parts[36] = 1
all error flags          = 0
convertTemperature(parts, 20)            = 24     (ambient, grill just lit)
convertTemperature(parts, 5/8/11/14/17)  = 960 -> None  (no probes plugged in)
```

That matches the physical state of the grill exactly, so the PBL index map is correct: the temperature block really does sit at 2/5/8/11/14/17 with the grill temperature at 20, and the status flags therefore cannot start before index 24.

**If the LBL frame has the same shape, then `LBL -> PBL` is not just a workaround, it is the correct layout** — which would explain why @entroversal's LG800BL works end to end against a PBL-board model.

### Minor, unrelated

Both routines contain:

```js
if (parts < 41) { return null; }   // LBL
if (parts < 44) { return null; }   // PBL
```

`parts` is an array, so this comparison is always false and the length guard never fires. Presumably meant to be `parts.length`. Harmless today since it fails open, but worth fixing if these routines are touched.

### Where a fix can realistically live

`grills.json` is regenerated daily from the vendor API (`update-grills.yml`), so the broken LBL `status_function` is upstream vendor data. Editing it in-repo would be overwritten on the next sync. That seems to argue for handling this with an alias/override, as this PR does at the config-flow level (or equivalently at the board level in pytboss), rather than trying to patch the routine itself.

Happy to capture and share more raw PBL frames (`FE0B` status and `FE0C` temperatures) in various states if reference data would help the tests.

dknowles2 (commented):
Nice refactor and fix! Converting the hardcoded `PBL2` check into `CONTROL_BOARD_ALIASES` makes control board mapping clean and extensible. Mapping `LBL` -> `PBL` correctly allows `LBL-*` discovered devices to match the `PBL` control board in `pytboss` rather than aborting with `unknown_grill`.

Code quality looks great and all CI checks are passing. Looks good to merge; David (`@dknowles2`) has final say.

trailro:
Follow-up, and I need to correct part of my previous comment.

I got hold of the actual PitBoss ESP firmware (it's Mongoose OS mJS source, not a binary) and it settles this from the vendor side. The firmware files are served publicly, no auth, at `https://firmware.dansonscorp.com/{version-with-dashes}/{file}` — e.g. `https://firmware.dansonscorp.com/0-2-3/init.js`. Versions currently up: 0.2.3, 0.4.0, 0.5.0, 0.5.5, 0.5.6, 0.5.7, 0.6.0.

### The vendor's own board table

Firmware **0.2.3** (`init.js`) carries this literal, before it was later moved into board-specific `platform_*.js` files:

```js
let powerStatusPos = {
  PBL: 24,
  PBC: 24,
  PBG: 24,
  PBM: 18,
  LFS: 21,
  LBL: 21,
  PBA: 27,
  PBB: 27,
  PBV: 24,
  PBP: 24,
};
```

and it is used as:

```js
if (pck.at(1) === 11) {
  let deviceType = deviceId.slice(0, 3);
  let pos = powerStatusPos[deviceType];
  if (pck.length > pos) {
    moduleIsOn = pck.at(pos) === 1;
  }
}
```

(For reference, current firmware 0.5.7 ships `platform_PBL.js` containing just `let powerStatusPos = 24;`, matching the table.)

### Correction to what I said earlier

I previously suggested the LBL routine's post-temperature indices looked like PBL's shifted down by 3 and were therefore wrong. **That was incorrect.** `LBL: 21` is the vendor's own value, and `grills.json`'s LBL routine reads `moduleIsOn: parts[21]`, so **the LBL flag block (21–40) is right**. Sorry for the noise.

### What is actually wrong in the LBL routine

Reconstructing the layout from the two confirmed anchors (PBL 24, LBL 21) and a 3-byte-per-temperature block:

| field | PBL | LBL |
|---|---|---|
| p1Target … p4Temp | 2, 5, 8, 11, 14 | 2, 5, 8, 11, 14 |
| smokerActTemp | 17 | *(absent)* |
| grill temperature | 20 | **17** |
| set/current discriminator | 23 | **20** |
| moduleIsOn | 24 | 21 |
| flags → recipeTime | 25 … 43 | 22 … 40 |

That 3-byte difference is exactly one temperature field, and it lines up with the length guards in `grills.json` (`< 44` for PBL, `< 41` for LBL) and with a real PBL frame I captured (45 bytes incl. the trailing `FF`), implying ~42 bytes for LBL.

So the LBL routine has two concrete bugs, both in the temperature section only:

1. `smokerActTemp: convertTemperature(parts, 14)` duplicates `p4Temp`'s index — LBL has no smoker-temp field, so this should be dropped.
2. `switch (parts[23])` with `convertTemperature(parts, 20)` should be `switch (parts[20])` with `convertTemperature(parts, 17)` — as written, the grill-temperature block (20–22) overlaps `moduleIsOn` at 21 and `err1` at 22, which is what makes the routine unusable.

### Caveat that matters for this PR

If a grill's frames really are LBL-layout, then aliasing `LBL -> PBL` would read `moduleIsOn` at 24 instead of 21 and misparse. Yet @entroversal reports an LG800BL working end to end against a PBL-board model, which suggests the Black Label line spans two hardware generations that both advertise as `LBL-*`: older units with the LBL layout (`powerStatusPos` 21) and newer ones actually carrying a PBL board (24).

If that's right, the alias is correct for newer units and wrong for older ones, and fixing the native LBL routine (2 lines, above) would cover both. @entroversal — if your LG800BL reports sane `moduleIsOn`/error flags through the PBL mapping, that's good evidence yours is the PBL-board generation.

Happy to share the downloaded firmware files or a raw PBL frame if either is useful.

dknowles2:
I think this should be properly resolved in https://github.com/dknowles2/pytboss/pull/491 but since I don't have this grill I can't confirm the actual data frames match up to my assumptions. I'll cut a new release and folks can test it out. If it's not working correctly, feel free to file an issue and I'll try and dig in more.

PR #277, opened 2026-06-26
feat: add options flow to override temperature unit (#113)

### LINKED ISSUE #113: Add option to specify temperature unit
### Checklist

- [x] I have filled out the template to the best of my ability.
- [x] This only contains 1 feature request (if you have multiple feature requests, open one feature request for each feature request).
- [x] This issue is not a duplicate feature request of [previous feature requests](https://github.com/dknowles2/ha-pitboss/issues?q=is%3Aissue+label%3A%22Feature+Request%22+).

### Is your feature request related to a problem? Please describe.

In Home Assistant we have the ability to set the unit of a sensor entity to whatever we need but the climate entity lacks this functionality. 

Image

Image

### Describe the solution you'd like

I'm requesting a setting for the integration that would force a specific temperature unit (C/F). We're strange in Canada as we use mix of metric and imperial measurements and for some reason imperial is the standard for cooking temperatures... I'd prefer to keep my Home Assistant globally set to Celsius as that's what we use to measure any other temperatures.

Thanks!

### Describe alternatives you've considered

I can change the global setting, but that is not ideal as other temperatures are measured in Celsius.

### Additional context

N/A

dknowles2:
## Problem

Users who cook in °F but have Home Assistant configured for °C (or vice versa) have no way to control which unit the PitBoss integration displays. HA's climate entity does not expose a per-entity unit toggle in the UI, and the only existing control is the `isFahrenheit` flag on the grill itself (configured through the PitBoss app).

See #113.

## Solution

Add an **Options flow** so users can explicitly override the temperature unit per-integration:

| Option | Behaviour |
|--------|-----------|
| **Auto** (default) | Use the unit reported by the grill (`isFahrenheit` flag) — existing behaviour |
| **Fahrenheit (°F)** | Always display in °F regardless of grill setting |
| **Celsius (°C)** | Always display in °C regardless of grill setting |

The override is respected by all temperature-bearing entities:
-  — climate entity temperature unit
-  — probe 1–4 sensor unit
-  — probe target number entity unit

Users can access this option via **Settings → Devices & Services → PitBoss → Configure**.

## Changes

- ****: Add  key and unit-name constants
- ****: Add  class and hook  on 
- ****: Check options override before auto-detecting from grill state
- ****: Check options override before auto-detecting from grill state
- ****: Check options override before auto-detecting from grill state
- ****: Add options step and  selector labels

Closes #113


trailro:
Reporting from production, as offered in #321 — I've been running this PR on a PB1600PS1 (PBL board) for about four weeks with `temperature_unit: celsius`.

**The options flow itself works well.** The option persists, the selector reads correctly, and it takes effect without a restart. No complaints there.

**But there's a bug worth fixing before this merges.** The PR changes the unit each entity *declares* and never converts the value:

```python
if unit_option == TEMPERATURE_UNIT_FAHRENHEIT:
    return UnitOfTemperature.FAHRENHEIT
if unit_option == TEMPERATURE_UNIT_CELSIUS:
    return UnitOfTemperature.CELSIUS
```

Every temperature in a state frame is in **the grill's own unit**, whichever that is. So when the override disagrees with the grill, the raw number gets relabelled rather than converted, and a grill running in Fahrenheit at 225 °F displays as "225 °C".

It's invisible as long as the two agree, which is why it survived my first weeks — my grill is in Celsius and so is the override. It became reachable for me only once I added a select entity for the grill's own unit, at which point flipping the grill to °F made every entity wrong.

Affected: `climate` (current + target), the probe temperature sensors, and the probe target numbers.

The fix I'm running is a single coordinator helper that converts from the grill's unit into whatever is being displayed, with all three platforms calling it instead of returning raw values. Same shape as the existing code, just with the conversion added. I can send it as a follow-up PR once this lands, or as a review suggestion here if you'd rather fold it in — whichever is less work for you.

One related note: writes have the mirror of this problem. Commands must always be expressed in the grill's unit, not the displayed one, so the same conversion is needed in the other direction on the way out.

trailro:
This introduces the first options flow in the repo, and I think it needs an update listener to work — otherwise changing the option has no effect until Home Assistant restarts.

`async_create_entry(title="", data=user_input)` stores the options, but nothing reloads the entry, and `__init__.py` is not touched by this PR. The usual pairing is:

```python
entry.async_on_unload(entry.add_update_listener(_async_options_updated))

async def _async_options_updated(hass: HomeAssistant, entry: ConfigEntry) -> None:
    await hass.config_entries.async_reload(entry.entry_id)
```

**Reading the option live is not enough here, and I checked rather than assumed.** Home Assistant pins an entity's displayed unit in the registry when it is first registered; a later change to `native_unit_of_measurement` is *converted into* that pinned unit rather than adopted.

I demonstrated it on the existing dynamic unit rather than on this PR's option — same mechanism, since both drive `native_unit_of_measurement`:

```
registered with isFahrenheit=True   →  unit_of_measurement=°F   state=165
changed to    isFahrenheit=False    →  unit_of_measurement=°F   state=165.0
```

The second reading has a native value of 74 °C. It is displayed as 165.0 °F — converted, not adopted. Applied to this PR: a user who picks Celsius keeps seeing Fahrenheit, and the only visible effect is that the number stops being a round setpoint.

Worth saying that a pinned display unit is the *right* behaviour for a grill that switches units on its own — the user should not have their dashboard flip because someone pressed a button on the panel. It is only wrong when the user is the one asking for the change, which is exactly what this PR adds.

**Also relevant to sequencing.** This is the first options flow, and two other things want one: adaptive polling's two intervals, and set 5's temperature-unit select from #321. Whichever lands first should own the flow and the listener, and the rest should extend it. I would rather rebase onto this than send a competing version — I have the polling half written and can hold it.

I run the listener above locally and it behaves as expected: changing an option reloads the entry, entities are rebuilt, and the new unit takes effect without a Home Assistant restart. Happy to send it as a small follow-up PR against this branch, or you can fold it in — whichever you prefer.

One thing it will not fix, and I mention it because it surprised me: reloading the entry does **not** reload edited Python. Module code is imported once, so anyone testing this by editing files still needs a restart. The listener helps users, not developers.

trailro:
I need to correct something I told you on this PR, and it changes the recommendation rather than refining it.

**I said Home Assistant pins an entity's displayed unit at registration.** That was wrong. I demonstrated the symptom correctly — the displayed unit does not follow `native_unit_of_measurement` — but I got the mechanism wrong, and the real one is worse for this PR.

**It is the unit system.** For an entity with `device_class: temperature`, Home Assistant converts the native value into the user's configured unit system, every time. Same entity, same code, only `hass.config.units` differs:

```
HA in US customary, grill reporting °F  ->  °F 165
HA in US customary, grill reporting °C  ->  °F 165.0      (native 74 °C)
HA in metric,       grill reporting °F  ->  °C 74.0       (native 165 °F)
HA in metric,       grill reporting °C  ->  °C 74
```

It is not pinned once and then stuck. It is re-derived from the unit system on every update, and the native unit only decides what gets converted *from*.

I also checked whether a reload rescues it — it does not. Reloading the config entry leaves the displayed unit exactly where it was, because there is nothing to rescue: the unit system still governs.

**So `CONF_TEMPERATURE_UNIT` cannot do what it is meant to do.** A user on a US-customary Home Assistant who sets `temperature_unit: celsius` still sees °F, because the override changes `native_unit_of_measurement`, and Home Assistant converts away from that into °F regardless. The option would appear to do nothing, which is a worse failure than not offering it.

And the update listener I recommended two comments up **would not have fixed this**. It makes the option take effect; it does not stop Home Assistant converting afterwards. I owe you that correction too — I suggested a fix for the wrong problem.

### What this means

Home Assistant already provides the setting: the user's unit system, plus a per-entity unit override in the entity settings dialog for entities that support one. This PR is trying to add a third layer underneath a system designed to be authoritative.

The only way an integration-level override could win is by dropping `device_class: temperature`, which turns off the conversion — and costs the device class on every temperature entity in the integration. That is a bad trade for a preference Home Assistant already exposes.

**So my recommendation is to close this**, and I would rather say that plainly than offer to rebase something I now believe cannot work. It also conflicts with `main` in five of its six files after #327, #333, #341 and #347, so a rebase is not cheap either.

If you would rather keep it, the honest shape is different from what is here: not an override of `native_unit_of_measurement`, but something that writes the per-entity unit into the entity registry — which is the layer the unit system respects.

### One consequence worth naming

I have been treating this PR as owning the first options flow, and holding two branches behind it. If it closes, the remote-start toggle in [#322](https://github.com/dknowles2/ha-pitboss/issues/322) introduces the first options flow instead — one field, built so anything else extends it. No collision either way; I just want the sequencing to be a decision rather than a race.

Separately: the grill-unit select in #347 is unaffected by any of this. It changes what the **grill** works in, which the board reports and the unit system then converts for display, so it does the thing it says regardless of anyone's unit system.

trailro:
Two precision fixes to my previous comment, since it argued from evidence and part of that evidence was sloppy. Neither changes the recommendation; the second strengthens it.

**The table said `°F 165.0` for a native 74 °C. The right number is 165.2** — 74 °C converts to 165.2 °F, and this repo's own `test_probe_native_value_and_unit` asserts exactly that for exactly this conversion. A transcription slip, not a different behaviour, but a comment correcting someone else's mechanism should get its own arithmetic right.

**"Home Assistant already provides the setting" was too broad.** I checked the core source rather than the UI this time: `sensor` and `number` both read a per-entity `unit_of_measurement` out of the entity registry options — that is the settings-dialog override, and it covers the probe sensors and the target numbers today, no integration code needed. **`climate` has no such option**, and the screenshots in #113 show precisely that gap — it is the entity the request is actually about.

That cuts against this PR harder, not softer: the climate card always displays in the unit system's unit, so *no* integration-level override — including a correctly-converting one — can change what the climate entity shows. The two entities this PR could plausibly serve already have the setting; the one entity #113 asks for cannot be served from here at all.

So for whoever lands on this from #113, the honest state of things:

- **Probe temperatures and probe targets**: settings dialog → unit, works today.
- **The climate card**: an HA core limitation, not this integration's. Workarounds are a template sensor over the climate attributes (template sensors accept the unit override) or a core feature request for a per-entity climate unit — that is where the fix belongs.

Recommendation unchanged: close, for the reasons in my previous comment plus the above.

dknowles2:
Closing this. The approach cannot work: Home Assistant re-derives a climate entity's displayed unit from the user's unit system on every update, so an integration-level override never reaches the user — it relabels without converting.

What people actually wanted from #113 is the grill-side unit control, and that landed in #347.

@trailro — this answers the question you raised on #322 and #344. Go ahead and introduce the options flow on `feat/start-grill`; there is nothing here for you to rebase onto or coordinate around. Building it as a single field that #351's polling intervals can extend later is the right shape.

Issue #321, opened 2026-08-01
[DONE] Roadmap: bug fixes and features from a fork running in production


trailro:
Hi — I've been running a fork of this integration on a **Pit Boss 1600 Pro Elite** (`PB1600PS1`, **PBL** control board, 4 probes incl. MPC) since early July. Along the way I fixed a number of bugs and added features, and validated everything against the real grill rather than just the protocol docs.

I'd like to contribute all of it back, but that's a lot of PRs and several touch `coordinator.py`, so I'd rather agree on shape and order with you first than dump branches on you. **Happy to reshape, merge, split or drop any of this** — this is a proposal, not a plan I'm attached to.

### Bug fixes

| | What | Notes |
|---|---|---|
| 1 | Typo `coordiantor` in `binary_sensor.py` | trivial |
| 2 | `GrillLight.available` ignores the connection | keeps reporting the last state after the grill goes away |
| 3 | Setup leaks the transport when the first refresh fails with anything other than `ConfigEntryNotReady` | socket + receive task survive; every retry strands another |
| 4 | A poll landing between the two status frames blanks all temperatures for a full cycle | the board clears `sc_11`/`sc_12` independently the moment it forwards a command to the MCU; pytboss omits the missing half's keys, so merging on absence (not on `None`) fixes it |
| 5 | Turning off sets **180 °C** on a grill running in Celsius | `spec.min_temp` is always Fahrenheit; affects both `climate` and `switch` |

### Correctness

| | What | Notes |
|---|---|---|
| 6 | Snap grill setpoints to `temp_increments` | the board silently ignores anything else |

While validating #6 I found the board **truncates** the °F→°C conversion where an obvious implementation rounds: asking for 88 °C (190 °F) lands on 87, and 99 °C (210 °F) lands on 98. Nine of the nineteen Celsius setpoints were wrong. Verified by sending each and reading back what the board stored.

### New entities and platforms

| | What |
|---|---|
| 7 | Fan + igniter binary sensors (`fanState` / `hotState`, already decoded but not exposed) |
| 8 | Connectivity binary sensor — a stable online/offline signal that stays available in order to report "disconnected" |
| 9 | Diagnostics from `Sys.GetInfo`: controller uptime and free memory, plus firmware version as a sensor and `sw_version` |
| 10 | Button platform: restart controller, keep WiFi awake |
| 11 | Select for the grill's own temperature unit (`set-celsius` / `set-fahrenheit`) |

### Probes

| | What |
[...]

PR #326, opened 2026-08-02
Keep fields missing from a partial state frame

trailro:
Set 2 of #321, first of three stacked on each other. Set 1 (#323, #324, #325) is merged — thanks for the quick turnaround.

The board answers with two independent frames, status (`sc_11`) and temperatures (`sc_12`), and clears both the moment it forwards a command to the MCU. A read landing in that window comes back with one frame but not the other, and pytboss omits the missing half's keys entirely — so every entity backed by them goes unknown until the next successful read.

On BLE the push path refills quickly enough that it is easy to miss. On a polling connection it lasts a full interval, and it happens after *every* command, including ones the integration issues itself.

Merging each frame onto the last known state fixes it. Only absent keys are carried over: a key present with a null value is a real reading — an unplugged probe — and still overwrites.

The merge also copies rather than storing the incoming dict, since pytboss hands every subscriber the same `StateDict` instance and keeps mutating it in place.

Three tests: the partial frame case, the null-reading case (a guard that we don't over-merge), and the pushed-update path.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -37,9 +37,32 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         await self.api.subscribe_state(self._on_state_update)
         await self._start_api()
 
+    def _merge_state(self, state: StateDict) -> StateDict:
+        """Fold a state frame onto the last known one.
+
+        The board answers with two independent frames, status (`sc_11`) and
+        temperatures (`sc_12`), and clears them as soon as it forwards a
+        command to the MCU. A read landing in that window returns one frame
+        without the other, and pytboss simply omits the missing half's keys,
+        so every entity backed by them would go unknown until the next
+        successful read.
+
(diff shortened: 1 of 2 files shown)


dknowles2 (commented):

LGTM! Excellent work.

### Summary of Review
- **Correctness & Design**: Merging incoming partial state frames onto `self.data` in `PitBossDataUpdateCoordinator` cleanly solves the issue where entities briefly went to `unknown` when the control board returns single status (`sc_11`) or temperature (`sc_12`) frames.
- **Edge Cases**: Correctly preserves explicit `None` values (e.g. unplugged probes) while copying state dicts to prevent mutating shared subscriber instances.
- **Resource Management**: Cleaning up sockets and stopping API connections on any setup exception in `__init__.py` prevents stranded tasks.
- **Code Quality & Testing**: Python cleanups (`LOGGER.warning`, `super().available`, typing fixes) accompanied by thorough unit test coverage across all affected components.
- **CI Status**: All workflow checks (build, mypy, ruff, HACS Validation, Hassfest Validation) are passing.

PR #327, opened 2026-08-02
Take the climate bounds from the board's setpoint list

trailro:
Set 2 of #321. Independent of #326 despite what I said there about stacking — GitHub won't let a PR from a fork use another fork branch as its base, so all three are against `main`. They touch different code and can merge in any order.

`min_temp` and `max_temp` came from `spec.min_temp`/`max_temp`, falling back to module constants when a model publishes neither. The board also publishes the exact list of setpoints it honours, which is a better source — `PB2180LK` has no `min_temp`, so the entity advertised a floor of 180 that the board does not impose, while its own list starts at **160**. That is the one existing expectation this changes, and I updated `test_min_temp` rather than working around it.

Celsius values are derived with `floor((F - 32) / 1.8)` rather than a rounding conversion, because that is what the boards which convert inside their own parsing routine do. Rounding produces values the panel never lands on: to the grill, 190 °F is 87 °C, not 88. Confirmed on my own grill by sending each value and reading back what it stored.

Where a model declares `celsius_temp_increment` that list wins — only PBV30DS/PBV30DX do today, and both are in `UNSUPPORTED_MODELS`, but the precedence seemed worth getting right.

The list is also published as an `allowed_setpoints` attribute, so a dashboard can offer only values that take effect instead of a free slider the board silently ignores.

Two notes:

- This deliberately does **not** change what gets sent. Snapping on the way out belongs in pytboss, and I've proposed it there (dknowles2/pytboss#506) since the current clamp is Fahrenheit-only and would undo any snapping done here.
- If the `accepted_setpoints()` from that issue lands, the coordinator method here becomes a one-line delegate. I kept the derivation local so this doesn't have to wait on a pytboss release.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -52,6 +52,21 @@ class GrillClimate(BaseEntity, ClimateEntity):
         self._attr_unique_id = f"{self.entity_description.key}_{entry_unique_id}"
 
     @property
+    def _accepted_setpoints(self) -> list[float]:
+        return self.coordinator.accepted_setpoints(self.temperature_unit)
+
+    @property
+    def extra_state_attributes(self) -> dict[str, list[float]] | None:
+        """Publish the setpoints the board accepts.
+
+        The board silently ignores anything else, so a dashboard offering a
+        free slider is misleading. This lets one offer only what works.
+        """
+        if accepted := self._accepted_setpoints:
+            return {"allowed_setpoints": accepted}
+        return None
+
+    @property
     def target_temperature_step(self) -> float:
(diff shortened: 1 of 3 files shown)


dknowles2 (commented):
LGTM! Code quality, correctness, and unit tests look great.

- Clean derivation of `accepted_setpoints` directly from the control board spec's setpoint increments.
- Correct handling of Celsius conversions matching panel behavior (`floor((F - 32) / 1.8)`) and model-specific Celsius lists when declared.
- `min_temp`, `max_temp`, and `extra_state_attributes["allowed_setpoints"]` properly reflect accepted values.
- All unit tests and CI checks are passing.

PR #328, opened 2026-08-02
Leave the setpoint alone when turning the grill off

trailro:
Set 2 of #321. Independent of #326 and #327; all three are against `main` (see the note in #327 about fork stacking).

Both turn-off paths wrote `spec.min_temp` to the grill afterwards. That value is always Fahrenheit, so on a grill running in Celsius it set the setpoint to **180 °C** — and even where the unit matched, it silently discarded whatever the user had dialled in. I lost my own setpoint to this a few times before working out where it went.

My first instinct was to fix the unit, but the setpoint has no effect while the grill is off and is exactly what the grill uses next time it is lit, so there is nothing to reset. Removing the write fixes the unit bug, stops the data loss, and is a smaller diff than getting the conversion right would be.

Happy to do it the other way if you'd rather keep the reset and just correct the unit — though that one can't be made correct on Celsius grills until the pytboss clamp is fixed.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -99,8 +99,10 @@ class GrillClimate(BaseEntity, ClimateEntity):
 
     async def async_turn_off(self) -> None:
         """Turn off the grill."""
+        # Deliberately does not touch the setpoint. It is the user's
+        # setting, it has no effect while the grill is off, and it is what
+        # the grill will use next time it is lit.
         await self.coordinator.api.turn_grill_off()
-        await self.async_set_temperature(temperature=self.coordinator.api.spec.min_temp)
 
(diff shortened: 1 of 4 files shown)


dknowles2 (commented):

## Code Review Summary

### LGTM
- **Correctness & User Experience**: Removing the explicit setpoint reset during `async_turn_off` in climate and switch platforms preserves the user's configured target temperature across grill power cycles and fixes a unit mismatch issue on Celsius grills (where `spec.min_temp` in Fahrenheit was written directly).
- **Code Quality**: Clean simplification in both `climate.py` and `switch.py`. Removed unused `DEFAULT_MIN_TEMP` import. Added helpful inline comments.
- **Tests & CI**: Updated `test_climate.py` and `test_switch.py` to assert `set_grill_temperature` is not called during turn-off. All tests pass and CI checks succeed.

PR #329, opened 2026-08-02
Expose the fan and igniter as binary sensors

trailro:
Set 3 of #321. Independent of #330; both are against `main`.

pytboss already decodes `fanState` and `hotState` out of the status frame, and the integration already surfaces `motorState` right next to them — these two just were never wired up.

Both are genuinely useful during a cook: the igniter tells you the grill is still lighting rather than holding temperature, and the fan running while the module is off is the shutdown cycle rather than a fault. `GrillClimate.hvac_action` already reads both for exactly that distinction, so the data was being used, just not shown.

Same shape as the existing `Auger` entity. The existing `test_creates_all_entities` covers them automatically; added one more for the reported state.

PR #330, opened 2026-08-02
Add a connectivity binary sensor

trailro:
Set 3 of #321. Independent of #329.

Every entity goes unavailable when the grill drops, which leaves nothing to build a "grill is offline" automation on — an entity that vanishes cannot report the state you want to trigger on.

This one deliberately overrides `available` to stay `True` and reports `api.is_connected()` instead. That is the opposite of the convention everywhere else in the integration, which is the point, so I've said why in the docstring rather than leaving it to look like an oversight.

I've been running it for about a month across all three transports — it only calls `is_connected()`, which every transport implements, so it behaves the same on BLE, cloud and a local connection.

One known limitation, not addressed here: if the grill is unreachable at startup the whole config entry is still in setup-retry, so this sensor is unavailable along with everything else. It works for connected-then-dropped, which is the case worth automating on.

### DIFF
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -117,10 +117,11 @@ async def async_setup_entry(
 ):
     """Setup binary_sensor platform."""
     coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
-    entities: list[BinarySensor] = []
+    entities: list[BinarySensorEntity] = []
     assert entry.unique_id is not None
     for entity_description in ENTITY_DESCRIPTIONS:
         entities.append(BinarySensor(coordinator, entry.unique_id, entity_description))
+    entities.append(ConnectivitySensor(coordinator, entry.unique_id))
     async_add_entities(entities)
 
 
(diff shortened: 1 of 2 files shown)


dknowles2 (commented):

Reviewed the diff and test additions for this connectivity binary sensor.

**Code quality:** Clean, small, well-scoped change. The docstring on `ConnectivitySensor` explains upfront *why* `available` is deliberately overridden to always return `True` — that is a nice touch given it's the opposite of the convention used by every other entity in this integration, so future readers won't mistake it for a bug.

**Correctness:** `is_on` guards on `self.coordinator.api` before calling `is_connected()`, avoiding an `AttributeError` if the coordinator hasn't picked up the API client yet. Behavior matches the PR description (author reports running it ~1 month across BLE/cloud/local transports).

**Tests:** `test_connectivity_reports_offline_instead_of_vanishing` covers both the online → offline transition and correctly asserts the entity state itself flips to `off` rather than going `unavailable`, which is the whole point of the sensor. Good coverage for the core behavior change.

**CI:** All checks (HACS validation, hassfest, build, mypy, ruff) are passing.

**Blocker:** GitHub reports this PR's merge state as `CONFLICTING`/`DIRTY` against the current base branch, so it cannot be merged as-is despite CI passing on the PR branch. @trailro — could you rebase/merge in the latest `main` and resolve the conflict?

The known limitation called out in the PR body (sensor unavailable during initial setup-retry when the grill is unreachable at startup) is reasonable to leave out of scope here, as noted.

Once the merge conflict is resolved, this looks ready for @dknowles2 to merge.

Recommendation: Needs Manual Intervention - please rebase onto main to resolve the reported merge conflict before this can be merged.

trailro:
Rebased onto `main` — the conflict was #329 landing in the meantime. Both PRs appended a test to the end of `tests/test_binary_sensor.py`, so it was purely additive; both tests are kept and the file now has `test_fan_and_igniter_report_their_state` followed by `test_connectivity_reports_offline_instead_of_vanishing`.

`mergeable` is back to `MERGEABLE`/`CLEAN` and all five checks pass on the rebased branch.

Thanks for the thorough read — and good catch on the ordering trap generally. The remaining sets have the same shape, so I'll rebase before raising each one rather than after, now that PRs are landing fast enough for base to move underneath them.

PR #331, opened 2026-08-02
Add a firmware version sensor

trailro:
Set 4 of #321. Independent of #332 and #333 — verified they merge cleanly against each other in any order.

pytboss exposes `get_firmware_version()` and the device page has a `sw_version` field for exactly this, but neither was used, so the device showed no firmware at all.

Read once during coordinator setup and surfaced both as a diagnostic sensor and on the device itself. The read is best-effort: a grill that does not answer simply gets no sensor rather than failing setup, which is covered by a test.

Unlike the other entities this one does not follow the connection — the value is read once and cannot change while the grill is away, so it stays readable rather than going unavailable.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -125,6 +125,9 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         await pitboss.stop()
         raise
 
+    if coordinator.firmware_version:
+        coordinator.device_info["sw_version"] = coordinator.firmware_version
+
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
     return True
 
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -20,6 +20,7 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
     config_entry: ConfigEntry
     device_info: DeviceInfo
     api: PitBoss
+    firmware_version: str | None = None
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


dknowles2 (commented):

LGTM. Small, focused addition: surfaces firmware version as both a diagnostic sensor and the device's `sw_version`, using an API pytboss already exposes but the integration wasn't calling. Good details:

- Read is best-effort at coordinator setup (`except Exception` + debug log), so a grill that doesn't answer just gets no sensor rather than failing integration setup — covered by a dedicated test.
- Correctly treated as read-once/static rather than following the connection state, since the value can't change while the grill is unreachable — matches the PR's stated reasoning and is a sensible deviation from the pattern used by the other (connection-following) sensors in this integration.
- Test coverage for both the happy path (sensor value + device `sw_version`) and the failure path (no sensor created).

dknowles2 (commented):

LGTM. Small, well-scoped addition that surfaces `pytboss.get_firmware_version()` — an API that already existed but wasn't being used — as both a diagnostic sensor and the device's `sw_version`. Good details:

- Read is best-effort at coordinator setup (`except Exception` + debug log), so a grill that doesn't answer just gets no sensor rather than failing integration setup — covered by a dedicated test (`test_no_firmware_sensor_when_the_grill_does_not_answer`).
- Correctly treated as read-once/static rather than following connection state, since firmware version can't change while the grill is unreachable — a sensible, well-documented deviation from the connection-following pattern used elsewhere in this integration (`available` explicitly checks `firmware_version is not None`, not `super().available`).
- Test coverage for both the happy path (sensor value + device `sw_version`) and the failure path (no sensor created).
- Squarely within this repo's boundary per REVIEW.md — coordinator/entity behavior, no protocol/parsing logic added.

CI is fully green (build, Hassfest, HACS, mypy, ruff) and mergeable state is MERGEABLE/clean.

PR #332, opened 2026-08-02
Add controller uptime and free memory diagnostics

trailro:
Set 4 of #321. Independent of #331 and #333.

The control board answers `Sys.GetInfo` with its own uptime and heap, reachable through the public `api.config.get_info()`.

Uptime is the one I actually wanted: a WiFi module that quietly reboots takes the connection with it, and without this there is nothing in Home Assistant that shows it happened. I lost about 32 minutes of a cook to exactly that before adding it — the uptime had dropped from ~47 hours to 1095 seconds, which is the only reason I know what happened.

Read at most once a minute rather than on every poll, since neither value changes fast. Both go unavailable with the connection so they cannot present a stale reading as current.

### DIFF
--- a/custom_components/pitboss/const.py
+++ b/custom_components/pitboss/const.py
@@ -9,6 +9,8 @@ NAME = "PitBoss"
 DOMAIN = "pitboss"
 MANUFACTURER = NAME
 PING_INTERVAL = timedelta(seconds=30)
+# Diagnostics change slowly; no need to read them at the poll rate.
+SYS_INFO_INTERVAL = 60.0
 PROTOCOL_WSS = "wss"
 PROTOCOL_BLE = "ble"
 ALL_PROTOCOLS = (PROTOCOL_BLE, PROTOCOL_WSS)
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -1,6 +1,7 @@
 """DataUpdateCoordinator for PitBoss."""
 
 from math import floor
+from time import monotonic
 
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import UnitOfTemperature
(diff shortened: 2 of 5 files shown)


dknowles2 (commented):

LGTM. Useful diagnostic addition motivated by a real incident (WiFi module reboot silently dropping the connection). Implementation is clean:

- Throttled to once a minute via `SYS_INFO_INTERVAL`/`monotonic()` rather than polling on every update — sensible given these values change slowly, and there's a test asserting `get_info` isn't re-awaited on back-to-back updates.
- The fetch is best-effort (`except Exception` + debug log) so a failure here can't break the coordinator's main update loop.
- Both sensors correctly go `unavailable` when the connection drops (`available` checks `super().available` combined with key presence in `sys_info`), avoiding stale readings — the stated goal.
- Reasonable device/unit classes (`DURATION`/seconds, `DATA_SIZE`/bytes with suggested KB display) and diagnostic entity category.

dknowles2:
@trailro Looks like there are merge conflicts.

dknowles2 (commented):

LGTM. Useful diagnostic addition motivated by a real incident (WiFi module reboot silently dropping the connection, costing ~32 minutes of a cook before this existed). Implementation is clean:

- Throttled to once a minute via `SYS_INFO_INTERVAL`/`monotonic()` rather than polling on every update — sensible given these values change slowly, and there's a dedicated test asserting `get_info` isn't re-awaited on back-to-back updates.
- The fetch is best-effort (`except Exception` + debug log) so a failure here can't break the coordinator's main update loop.
- Both sensors correctly go `unavailable` when the connection drops (`available` checks `super().available` combined with key presence in `sys_info`), avoiding stale readings — the stated goal.
- Reasonable device/unit classes (`DURATION`/seconds, `DATA_SIZE`/bytes with suggested KB display) and diagnostic entity category.
- The earlier merge conflict has been resolved (branch was rebased/updated after David's "merge conflicts" comment) and CI is now fully green (HACS, Hassfest, build, mypy, ruff all passing).

trailro:
Rebased — #331 landing first put both changes in the same two spots in `sensor.py` (the import block and the entity list), plus adjacent test additions.

Worth flagging how I got there: my first attempt resolved the conflict mechanically, keeping both sides of every hunk, and produced a file that didn't parse — the hunks weren't whole blocks. I threw that away and rebuilt the branch from the merged main instead, re-applying the change against the new code rather than reconciling two versions of it. Clean now, 84 tests, all checks green.

PR #341, opened 2026-08-02
Give every probe a target

trailro:
Set 6 of #321, second of three. Builds on #333, and on [pytboss#527](https://github.com/dknowles2/pytboss/pull/527), now released in `2026.8.3`.

The grill acts on one probe — the control probe — and that is the only one with a board command. Home Assistant therefore offers a target on probes 1 and 2 at most, and only on boards declaring `set-probe-N-temperature`. Across the catalogue that is **42 of 137 models for probe 1 and 26 for probe 2, and nothing at all for probes 3 or 4**. A four-probe grill gets two targets; on 95 models it gets none.

`pytboss.set_probe_target()` routes around that: the board command where one exists, the grill's virtual data store otherwise — which is what the vendor's app does. This PR gives every probe an entity and lets the library decide the route.

**This is the split version you asked for.** The protocol half — the `p1T`..`p4T` key naming, the Fahrenheit convention, the wholesale-assign merge, the `psw` strip, the power-off wipe, the `moduleIsOn` gate — moved to pytboss#527 and is not here. What is left is the Home Assistant half:

- an entity per probe, up to `meat_probes`
- `RestoreNumber`, converting if the grill's unit changed while HA was down
- the port naming from #333
- writing a held target to the grill when it comes back on

For scale: the original was `+378/−61` across four files; this is `+283/−63`, and the coordinator loses about a hundred lines.

**The three-way precedence is preserved, split along the same line.** pytboss owns board-reported over scratchpad — both are things the grill told us. Home Assistant keeps the third rung, a target we set and restored, because that one exists only because HA outlives the grill's store, which is wiped at power-off. Board-reported is read from `self.data[f"p{n}Target"]` directly rather than through the library, since the key is already in `StateDict` and reading it here means a target the grill announces shows on the push instead of at the next poll.

**Three behaviour changes, unchanged from the original and all previously agreed:**

1. Grills declaring no probe command now get target entities where they previously got none. That is the point of the change, but it is new entities on existing installs.
2. The target is no longer unavailable when the probe is unplugged — the grill accepts a target for an empty port, so it can be dialled in before the probe goes in.
3. `p2Target` entities appear on grills without a `set-probe-2-temperature` command.

**Verified against the released `2026.8.3` rather than an overlay**, on current `main` with both pins already at that version:

```
97 tests passed
ruff check   — all checks passed (with .ruff.toml from #343)
ruff format  — 24 files already formatted
mypy         — no issues in 24 source files
```

Squashed to one commit, since the second one existed only to undo the first.

Two smaller things worth recording, both caught by tests during the rewrite. `_async_refresh_probe_targets` was mutating the dict `get_probe_targets()` returned, so seeding ran once and then silently believed the grill already had the target — visible only on the second power cycle. And `native_value` returning `float` changed the displayed state from `165` to `165.0`; it returns the int.

The `native_unit_of_measurement` interaction I flagged before is untouched and still belongs with set 5: Home Assistant fixes an entity's display unit at registration, so a dynamic unit is converted into the pinned one rather than adopted. I noted the same thing on #277, where it decides whether the option has any effect.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -9,7 +9,12 @@ from homeassistant.core import HomeAssistant
 from homeassistant.helpers.device_registry import DeviceInfo
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator, UpdateFailed
 from pytboss.api import PitBoss
-from pytboss.exceptions import GrillUnavailable, NotConnectedError, RPCError
+from pytboss.exceptions import (
+    GrillUnavailable,
+    NotConnectedError,
+    RPCError,
+    UnsupportedOperation,
+)
 from pytboss.grills import StateDict
 
 from .const import DOMAIN, LOGGER, PING_INTERVAL, SYS_INFO_INTERVAL
@@ -39,6 +44,11 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         # Latest Sys.GetInfo payload from the control board.
         self.sys_info: dict = {}
         self._sys_info_at = 0.0
+        # What the grill is holding, and what we are holding for it. See
+        # `probe_target` for why both exist.
+        self.probe_targets: dict[int, int] = {}
+        self.restored_targets: dict[int, int] = {}
+        self._targets_seeded = False
 
     def accepted_setpoints(self, unit: str) -> list[float]:
         """Grill setpoints the control board honours, expressed in `unit`.
@@ -61,6 +71,69 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
     def has_mpc(self) -> bool:
         """Whether the grill has a meat probe control port."""
         return self.api.spec.has_mpc
+
+    @property
+    def grill_unit(self) -> str:
+        """The unit the grill is currently working in."""
+        if (data := self.data) and not data.get("isFahrenheit"):
+            return UnitOfTemperature.CELSIUS
+        return UnitOfTemperature.FAHRENHEIT
+
+    def probe_target(self, probe_number: int) -> int | None:
+        """This probe's target, in the grill's own unit.
+
+        Board-reported first: `pNTarget` is already in the state we hold, so
+        reading it here means a target the grill announces shows immediately
+        rather than at the next poll. Then what pytboss resolved from the
+        grill's store. Then ours -- a target we set and restored across a
+        restart, which exists only because Home Assistant outlives that store.
+        """
+        reported = (self.data or {}).get(f"p{probe_number}Target")
+        if isinstance(reported, (int, float)):
+            return int(reported)
+        if (target := self.probe_targets.get(probe_number)) is not None:
+            return target
+        return self.restored_targets.get(probe_number)
+
+    async def async_set_probe_target(self, probe_number: int, temp: int) -> None:
+        """Set a probe's target, keeping it if the grill cannot take it yet."""
+        self.restored_targets[probe_number] = temp
+        try:
+            await self.api.set_probe_target(probe_number, temp)
+        except UnsupportedOperation:
+            # The grill is off: its store rejects writes and is wiped anyway.
+            # Held here and written when it comes on.
+            return
+        self.probe_targets[probe_number] = temp
+
+    async def _async_refresh_probe_targets(self, state: StateDict) -> None:
+        """Track the targets the grill is holding. Never fatal."""
+        if not state.get("moduleIsOn"):
+            self.probe_targets = {}
+            self._targets_seeded = False
+            return
+        try:
+            # Copied: we add to this below, and it is not ours to mutate.
+            self.probe_targets = dict(await self.api.get_probe_targets())
+        except Exception as ex:  # noqa: BLE001
+            self.logger.debug("Could not fetch the probe targets: %s", ex)
+            return
+        if self._targets_seeded:
+            return
+        self._targets_seeded = True
+        # The grill has just come on, so anything we are holding that it does
+        # not know about goes over now. That is what makes setting a target on
+        # a cold grill mean anything. A target the grill already has was set
+        # elsewhere and wins.
+        for probe_number, temp in self.restored_targets.items():
+            if probe_number in self.probe_targets:
+                continue
+            try:
+                await self.api.set_probe_target(probe_number, temp)
+            except Exception as ex:  # noqa: BLE001
+                self.logger.debug("Could not seed a probe target: %s", ex)
+            else:
+                self.probe_targets[probe_number] = temp
 
     async def _async_setup(self) -> None:
         """Set up the coordinator."""
@@ -141,7 +214,9 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         # Relying solely on push notifications means sensors can go stale after
         # a reconnect if push notifications stop being delivered.
         try:
-            return self._merge_state(await self.api.get_state())
+            state = self._merge_state(await self.api.get_state())
+            await self._async_refresh_probe_targets(state)
+            return state
         except NotConnectedError as ex:
             raise UpdateFailed("Grill not connected") from ex
         except RPCError as ex:
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -106,6 +106,21 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
             return
         self.probe_targets[probe_number] = temp
 
+    def note_restored_target(self, probe_number: int, temp: int) -> None:
+        """Take a target restored from a previous run.
+
+        Re-arms seeding when the grill does not already hold this probe's
+        target. Entities are added after the coordinator's first refresh, so
+        a grill that power-cycled while Home Assistant was down has already
+        been marked seeded -- against an empty set -- by the time a restored
+        value arrives. Without this the target would show in Home Assistant
+        and never reach the grill, which is the one case the restore exists
+        for.
+        """
+        self.restored_targets[probe_number] = temp
+        if probe_number not in self.probe_targets:
+            self._targets_seeded = False
+
     async def _async_refresh_probe_targets(self, state: StateDict) -> None:
         """Track the targets the grill is holding. Never fatal."""
         if not state.get("moduleIsOn"):
--- a/custom_components/pitboss/number.py
+++ b/custom_components/pitboss/number.py
@@ -1,17 +1,15 @@
 """Number platform for pitboss."""
 
-from collections.abc import Awaitable, Callable
 from dataclasses import dataclass
 from typing import Literal
 
-from homeassistant.components.number import NumberEntity, NumberEntityDescription
+from homeassistant.components.number import NumberEntityDescription, RestoreNumber
 from homeassistant.components.number.const import NumberDeviceClass
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import UnitOfTemperature
 from homeassistant.core import HomeAssistant
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.util.unit_conversion import TemperatureConverter
-from pytboss.api import PitBoss
 
 from .const import (
     DEFAULT_PROBE_CELSIUS_STEP,
@@ -27,50 +25,37 @@ from .entity import BaseEntity
 
 @dataclass(frozen=True, kw_only=True)
 class PitBossNumberEntityDescription(NumberEntityDescription):
-    key: Literal["p1Target", "p2Target"]
-    probe_number: Literal[1, 2]
-    set_fn: Callable[[PitBoss], Callable[[int], Awaitable[dict]]]
+    key: Literal["p1Target", "p2Target", "p3Target", "p4Target"]
+    probe_number: Literal[1, 2, 3, 4]
     device_class: NumberDeviceClass = NumberDeviceClass.TEMPERATURE
     icon: str = "mdi:thermometer"
-    matching_probe_key: Literal["p1Temp", "p2Temp"]
 
 
-PROBE_1_DESCRIPTION = PitBossNumberEntityDescription(
-    key="p1Target",
-    probe_number=1,
-    set_fn=lambda api: api.set_probe_temperature,
-    matching_probe_key="p1Temp",
-)
-PROBE_2_DESCRIPTION = PitBossNumberEntityDescription(
-    key="p2Target",
-    probe_number=2,
-    set_fn=lambda api: api.set_probe_2_temperature,
-    matching_probe_key="p2Temp",
+PROBE_DESCRIPTIONS = (
+    PitBossNumberEntityDescription(key="p1Target", probe_number=1),
+    PitBossNumberEntityDescription(key="p2Target", probe_number=2),
+    PitBossNumberEntityDescription(key="p3Target", probe_number=3),
+    PitBossNumberEntityDescription(key="p4Target", probe_number=4),
 )
 
 
 async def async_setup_entry(
     hass: HomeAssistant, entry: ConfigEntry, async_add_devices: AddEntitiesCallback
 ):
-    """Setup number platformm."""
+    """Setup number platform."""
     coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
     assert entry.unique_id is not None
-    entities: list[TargetProbeTemperature] = []
-    if "set-probe-1-temperature" in (
-        available_commands := coordinator.api.spec.control_board.commands
-    ):
-        entities.append(
-            TargetProbeTemperature(coordinator, entry.unique_id, PROBE_1_DESCRIPTION)
-        )
-    if "set-probe-2-temperature" in available_commands:
-        entities.append(
-            TargetProbeTemperature(coordinator, entry.unique_id, PROBE_2_DESCRIPTION)
-        )
+    probe_count = coordinator.api.spec.meat_probes or 1
+    entities = [
+        TargetProbeTemperature(coordinator, entry.unique_id, description)
+        for description in PROBE_DESCRIPTIONS
+        if description.probe_number <= probe_count
+    ]
     if entities:
         async_add_devices(entities)
 
 
-class TargetProbeTemperature(BaseEntity, NumberEntity):
+class TargetProbeTemperature(BaseEntity, RestoreNumber):
     """PitBoss target probe temperature class."""
 
     def __init__(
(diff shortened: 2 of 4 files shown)


custom_components/pitboss/coordinator.py
```diff
@@ -62,6 +68,132 @@ def has_mpc(self) -> bool:
...
+
+        The firmware assigns the payload wholesale (`vData = params`), so
+        anything not sent back is lost -- the vendor's app merges client-side
+        too. `psw` is dropped because authentication injects it into the same
+        object and it would otherwise be stored as data.
+        """
+        payload = {k: v for k, v in self.virtual_data.items() if k != "psw"}
+        payload.update(updates)
+        await self.api.set_virtual_data(payload)
+        self.virtual_data = payload
```
dknowles2: **[PROBLEM]** This captures the injected `psw` into `self.virtual_data`.

`set_virtual_data` passes `payload` to `_authenticate`, which mutates it in place — `params["psw"] = encode(...).hex()` on the caller's dict — so by the time this line runs, `payload` has the encoded password in it. `self.virtual_data` then holds it until the next write strips it out again at line 132.

Only on password-protected grills, and there is no diagnostics platform in this integration today, so nothing is currently exposing it. But it is a credential-derived value sitting in a public attribute on the coordinator, and "nothing dumps coordinator state yet" is a thin guarantee — a diagnostics platform is the obvious next thing this repo grows.

The local fix is to snapshot before the call:

```python
payload = {k: v for k, v in self.virtual_data.items() if k != "psw"}
payload.update(updates)
await self.api.set_virtual_data(dict(payload))
self.virtual_data = payload
```

The real fix is upstream, per the summary: `_authenticate` should not mutate its argument, and `set_virtual_data` should not be able to write a password into the grill's scratchpad. If the routing moves to pytboss this disappears on its own.

custom_components/pitboss/coordinator.py
```diff
@@ -62,6 +68,132 @@ def has_mpc(self) -> bool:
...
+
+        The board reports a target of its own only for the control probe --
+        `p1Target` on every model, `p2Target` on 26 of them -- so for the rest
+        the scratchpad, and failing that our own copy, is the only source.
+        """
+        if (data := self.data) is not None:
+            reported = data.get(f"p{probe_number}Target")
+            if isinstance(reported, (int, float)):
+                return int(reported)
+        raw = self.virtual_data.get(f"p{probe_number}T")
```
dknowles2: This three-way precedence — board-reported, then scratchpad, then our own copy — is the part I would most want to keep exactly as written when this moves upstream. It is the right order, and the reasoning in the docstring is what makes it reviewable.

Noting it so it does not get lost in the restructuring: whatever the pytboss API ends up looking like, this fallback chain is the behaviour worth preserving verbatim.

dknowles2 (changes requested):
The feature is right and I want it. The investigation behind it is the best-evidenced thing you have sent — reading `PB.GetVirtualData` / `PB.SetVirtualData` across all seven mJS firmware versions, establishing the power-off wipe empirically, and being explicit that you can confirm presence but not behaviour on the ESP-IDF generation is exactly the standard I would want and rarely get.

**And your catalogue numbers are right where mine were wrong.** I counted 41 and 27 from `grills.json` and was about to query it. I was wrong twice over: the raw JSON misses `_COMMAND_SLUG_OVERRIDES` (the vendor's `set-prove-1-temperature` typo, corrected at parse time), and it counts the ten shadow-variant keys as models. Through `get_grills()` it is exactly your 42 and 26 of 137, and no board declares anything for probes 3 or 4 — the only probe slugs in the whole catalogue are `set-probe-1-temperature` and `set-probe-2-temperature`. Recording that so nobody re-derives it from the JSON and gets my answer.

What I am asking for is a split, not a rethink.

## The protocol half belongs in pytboss

`coordinator.py` gains about a hundred lines that encode facts about the firmware:

* the scratchpad keys are `p1T`..`p4T`
* values are always Fahrenheit, whatever the grill is displaying
* the firmware assigns the payload wholesale, so writes must merge client-side
* `psw` has to be stripped before writing
* it is wiped at power-off
* it only accepts writes while the grill is on

Every one of those is true with no Home Assistant anywhere near it. That is the test [REVIEW.md](https://github.com/dknowles2/ha-pitboss/blob/main/REVIEW.md) applies, and this is the largest amount of protocol knowledge this repo has carried since the `PBL2` remap. Different in kind — that one was wrong and this one works — but the same reason it is a problem: the next person who needs probe targets from Python has to reimplement all six facts, and the first they will learn about the power-off wipe is when a target silently vanishes.

**The `psw` strip is the tell, and it is our bug, not yours.** `PitBoss._authenticate` does `params["psw"] = ...` on the dict it was handed and returns the same object, and `set_virtual_data(data)` passes the caller's dict straight into it. So pytboss injects the encoded password into the payload, the firmware stores it as data, and this PR cleans up after us. Home Assistant core's rule is the one I would cite: *integrations should not implement workarounds for limitations in libraries; the library should be fixed*. `set_virtual_data` should not be able to write a credential into the grill's scratchpad, and no consumer should have to know that it can.

**Also, the private-API claim in the description is not quite right.** Both `set_virtual_data` and `get_virtual_data` carry `:meta private:` in their docstrings — they are excluded from the generated docs. Not underscore-prefixed, so an easy thing to miss, but they are marked private rather than public.

## What I would do instead

A pytboss PR first, adding something like:

```python
async def set_probe_target(self, probe_number: int, temp: int) -> None: ...
async def get_probe_targets(self) -> dict[int, int]: ...
```

owning the key naming, the Fahrenheit convention, the merge, the `psw` fix, and the routing between `set-probe-N-temperature` and the scratchpad — since which route a board supports is a fact about the board. Then this PR keeps what is genuinely ours and is already good: the entity per probe, `RestoreNumber` and the unit conversion on restore, the naming, and the decision to offer a target on a probe the grill has no opinion about.

I realise that is a bigger ask than the change deserves on its own, and that it blocks you behind a release. I still think it is right, and I would rather say so before it lands than carry it.

## The behaviour changes

All three are fine by me, and thank you for listing them rather than letting me find them. Dropping the unplugged-probe gate is a straightforward improvement — the entity vanishing exactly when you want to dial a target in was the wrong behaviour. New target entities on existing installs is the point of the PR.

The `native_unit_of_measurement` interaction you flagged and did not fix: agreed, that belongs with the temperature-unit work in set 5. Moving the test down to the coordinator rather than papering over it was the right call.

One smaller thing noted inline.

CI is green and I have labelled this `enhancement`.

trailro:
Agreed, and doing it — [dknowles2/pytboss#527](https://github.com/dknowles2/pytboss/pull/527) is up with `set_probe_target` / `get_probe_targets` owning the routing, the key naming, the Fahrenheit convention, the merge, and the `psw` fix.

Three corrections to things I said, all in your favour.

**The private-API claim was wrong.** `set_virtual_data` and `get_virtual_data` both carry `:meta private:`. I checked for underscores and stopped there, so "no private pytboss API is used" was not true of this PR. That is a worse error than it looks: the whole argument for the change being self-contained rested on it.

**The `psw` capture is real and I confirmed it rather than taking your word.** `_authenticate` does `params["psw"] = ...` on the caller's dict and returns the same object, so `self.virtual_data` holds the credential until the next write strips it. Your point about "nothing dumps coordinator state yet" being a thin guarantee is the right way to weigh it.

**And there is a part of it neither of us can fix.** Writing my test for the upstream fix, I asserted the payload reaching the grill was clean — and it failed on the password fixture. `PB.SetVirtualData` carries authentication in the same params object the firmware assigns wholesale, so an authenticated write necessarily leaves `psw` in the scratchpad *on the device*. The library can stop it reaching callers, which #527 does on both the write and read sides, but it cannot keep it off the grill. The test now records that rather than asserting something untrue.

Which sharpens your argument rather than weakening it: if a consumer has to know that reading virtual data hands back a credential, that is precisely the knowledge that should not be spread across consumers.

**On the split.** #527 takes the protocol half. This PR keeps the entity per probe, `RestoreNumber` with the unit conversion on restore, the naming, and offering a target on a probe the grill has no opinion about.

The three-way precedence you asked me to preserve is preserved, and split along the same line: pytboss owns board-reported over scratchpad, since both are things the grill told us. Home Assistant keeps the third rung — our own restored copy — because that one only exists because HA restarts and the grill's store does not survive a power cycle. Same order, same reasoning, in the two places that own the two kinds of fact.

I have not restructured this PR yet, since it is now blocked behind a pytboss release either way. Two ways to go and I do not mind which:

1. Leave it as is until #527 lands and a release follows, then rewrite it on top.
2. Close it now and reopen the thin version when there is a tag, so it is not sitting here looking ready.

If you would rather it not carry the `psw` capture in the meantime, the one-line `dict(payload)` fix from your inline comment is easy — though it disappears entirely in the rewrite.

**On the catalogue numbers** — worth recording that the raw-JSON route misses `_COMMAND_SLUG_OVERRIDES`, since that is the vendor's `set-prove-1-temperature` typo being corrected at parse time. I only got 42 and 26 right because I counted through `get_grills()`, which was luck rather than judgement: I had been burned earlier today by counting `grills.json` entries as models and telling you there were 137 of them when the file has 147.

dknowles2:
**#527 is merged**, along with #525. So the protocol half is on `main` and the only thing between this PR and a rewrite is a pytboss release, which is David's to cut.

**Option 1 — leave it open.** Closing would lose this thread, and the PR is not misleading anyone: it is sitting at changes-requested with the reason on it. An open PR blocked on a release is a normal state; an issue-less closed one is harder to find again.

No need for the `dict(payload)` patch in the meantime. It only matters on a branch nobody is running, and it disappears in the rewrite — spending a push on a line you are about to delete is the sort of tidying that makes a diff harder to read later.

For the rewrite, what survives from here is what we agreed: the entity per probe, `RestoreNumber` with the unit conversion, the naming, and offering a target on a probe the grill has no opinion about. The routing, the key naming, the Fahrenheit convention, the merge and the power-off semantics are now `set_probe_target()` / `get_probe_targets()` upstream.

Your line on the three-way precedence is better than mine and I have adopted it: pytboss owns board-reported over scratchpad because both are things the grill told us, and this repo keeps the restored copy because that rung exists only since Home Assistant restarts and the grill's store does not survive a power cycle.

One thing worth carrying into the rewrite: `get_probe_targets()` omits probes with no target rather than mapping them to `None`, so the coordinator's fallback wants `targets.get(n)` rather than a membership test that treats a stored `None` as a value.

Also — you flagged that getting the catalogue numbers right was luck rather than judgement. Worth saying that the correction is now written down: counting through `get_grills()` rather than raw `grills.json` is the rule, because the raw file misses `_COMMAND_SLUG_OVERRIDES` and counts the ten shadow-variant keys as models. I got that wrong reviewing this PR and you did not, which is the direction that matters.

trailro:
Built the split version so this is not waiting on me when a release lands. It is on [`trailro:feat/probe-targets-thin`](https://github.com/trailro/ha-pitboss/compare/main...trailro:ha-pitboss:feat/probe-targets-thin) rather than pushed here, because pushing it would turn this PR's CI red for a reason nobody could act on.

**It cannot merge on a release alone**, which is worth stating plainly since "wait for the tag" would be the natural assumption. Three things have to happen first, and only the last is mine:

1. [pytboss#527](https://github.com/dknowles2/pytboss/pull/527) merges
2. a pytboss release is cut — `main` is ten commits past `2026.8.2` and none of the nine merged today are in a tag
3. `requirements.txt` **and** `manifest.json` both move, per your #522

Then this gets force-updated to the branch above and goes green.

**What the split actually removed:**

```
this PR   : 4 files changed, 378 insertions(+),  61 deletions(-)
split     : 4 files changed,  91 insertions(+), 188 deletions(-)
```

The coordinator loses `virtual_data`, `probe_command_slug`, `_async_write_virtual_data`, `_async_refresh_virtual_data` and both Fahrenheit converters — about a hundred lines, all of it the protocol knowledge you identified. What is left is the entity per probe, `RestoreNumber` with the unit conversion on restore, the naming, and the seeding policy.

**Verified against the proposed API, not assumed.** I installed pytboss from the #527 branch over the pinned one in the test container and ran the suite: 95 tests, ruff, ruff format and mypy clean. Without that overlay it fails exactly as you would expect — `Mock object has no attribute 'set_probe_target'` — which is the same trap as #339, caught before it could reach anyone.

**Two things came out of doing it, both improvements:**

The three-way precedence you asked me to preserve now reads better than it did. Board-reported comes from `self.data[f"p{n}Target"]` directly rather than through the library, because that key is already in `StateDict` and reading it here means a target the grill announces shows on the push instead of waiting for the next poll. Then what pytboss resolved from the store, then ours. Same order, and the middle rung is the only one that costs a round trip.

And a test caught a real bug I introduced in the rewrite: `_async_refresh_probe_targets` was mutating the dict `get_probe_targets()` returned. Since a mock returns the same object every call, seeding ran once and then silently believed the grill already had the target. It is `dict(...)` now. Mutating a value a library handed back was my mistake, and it is the sort that only shows up on the second power cycle.

**On the `psw` capture in the meantime** — it disappears entirely in the split, so I have not patched it here. If you would rather this PR not carry it while it waits, say so and I will push the one-line fix; otherwise I would rather not add a commit that is deleted by the next one.

Happy to close this and open the thin one fresh instead, if a PR that cannot merge for a while is worse than no PR. Your call — I have no preference beyond not leaving something misleading open.

trailro:
Re-verified the split version against everything that has landed since I built it, because three things moved underneath it.

**`main` here moved.** #342 and #343 landed. #343 matters more than it looks: it adds `.ruff.toml` and makes CI run `ruff format --check` as well as `ruff check`, so a branch that passed before #343 has not actually been checked against what CI now enforces. Rebased onto `3904d96`.

**`main` in pytboss moved** — #534 through #538 merged after I built this.

**And [pytboss#541](https://github.com/dknowles2/pytboss/pull/541) is open and touches `set_probe_target` directly.** It passes the grill's unit flag as the command routine's second argument, because on eleven boards the MCU always speaks Fahrenheit and the vendor's routine only converts when told the panel is in Celsius. That is a real behaviour change in the exact call this PR delegates to — but it changes what pytboss puts on the wire, not the signature, so nothing here needs to move. Which is the split working as intended: the unit convention is a protocol fact and it is being fixed on the protocol side.

Verified rather than reasoned: I overlaid pytboss `main` **with #541 applied** over the pinned install and ran the suite.

```
97 tests passed
ruff check      — all checks passed (with the new .ruff.toml)
ruff format     — 24 files already formatted
mypy            — no issues in 24 source files
```

Pushed at `96b81da` on [`feat/probe-targets-thin`](https://github.com/trailro/ha-pitboss/compare/main...trailro:ha-pitboss:feat/probe-targets-thin). Still not pushed here, for the same reason as before: it would turn this PR's CI red against the pinned `2026.8.2`.

**One thing I checked for my own sake and will record because it cuts against an assumption.** I expected #541's bug to be biting my grill, since it runs in Celsius. It is not: PBL is not among the eleven boards, and its `set-temperature` and `set-probe-1-temperature` routines take one parameter and ignore a second. So the boards this repo's tests are built on — PBV4PS2 is also PBL — are precisely the ones that cannot show the bug. Worth knowing before anyone reads green tests here as evidence about it either way.

Nothing else outstanding from me on this one. The chain is still: #541 and #530 land → a release → the two-pin bump → this gets force-updated to the branch above.

dknowles2:
Two updates and one correction, then I will leave this parked.

**#541 merged** about a minute before you posted — you were overlaying an open PR that had just landed, so your run was against what is now pytboss `main`.

**The count is nine boards a user can reach, not eleven.** I corrected this on #541 itself, repeating it here so the number does not travel: eleven is right for entries in `control_boards`, but `PBVA`, `PBX1` and `PBL2` have **zero grill models pointing at them**. 20 declared boards, 17 reachable via `get_grills()`. The fix is correct for all eleven and should stay that way, but no catalogued grill is affected on those two.

**Your PBL observation is right, and it is the more useful half of that comment.** I checked it independently rather than taking it:

```
PBV4PS2 -> board PBL
  set-temperature          (100)=FE0501010000FF  (100,False)=FE0501010000FF  affected: False
  set-probe-1-temperature  (100)=FE0502010000FF  (100,False)=FE0502010000FF  affected: False
```

So the board this repo's fixtures are built on is one of the eight that cannot exhibit the bug at all. Green tests here were never going to say anything about it in either direction, and that is worth having written down somewhere permanent — it is the kind of thing that gets mistaken for coverage later. Same shape as the point I made on #530 about a button whose RPC is a no-op on the transport you actually use: the test bed and the failure mode do not overlap.

On the split: agreed, and your reasoning is the reason I asked for it. #541 changed what goes on the wire without changing the signature this PR delegates to, so the protocol fix landed on the protocol side and nothing here had to move. That is the boundary working.

Still parked on the release. Nothing needed from you.

dknowles2:
**Unblocked — push whenever you are ready.**

pytboss [2026.8.3](https://github.com/dknowles2/pytboss/releases/tag/2026.8.3) is out, and #345 moved both pins to it: `requirements.txt` and `custom_components/pitboss/manifest.json`. `main` is at `f3b2e75`.

I verified the API your thin version depends on against the installed release rather than a checkout — `set_probe_target`, `get_probe_targets` and `probe_target_command` are all present in 2026.8.3, along with the fourteen methods this integration already calls, signatures unchanged.

So a force-push of `feat/probe-targets-thin` should go green here rather than red against the old pin, which was the whole reason you were holding it.

Two things that landed after you last rebased, neither of which should move anything for you:

* **#530 merged** — `reboot()` and `request_fast_updates()` (renamed from the RPC's misleading `PB.WiFiAwakeWDT`; David confirmed the name). Worth knowing for the downstream button you mentioned: it is cloud-only *and* only does anything while the grill is on, so on Bluetooth or local it is a no-op in both directions.
* **#542** — I filed the `Transport.send_command -> dict` annotation against ourselves as promised, so it is tracked and out of your way.

I will review the rewrite fresh when it lands. Same thing I said before still applies: `set_probe_target()` / `get_probe_targets()`, coordinator.py losing the scratchpad semantics, and the fallback on `targets.get(n)`.

One note carried over from the release, since it touches this PR's neighbourhood: #541 changed what temperature commands put on the wire for Celsius grills, and **neither repo's tests can demonstrate it** — `PBV4PS2` is a PBL board, one of the eight whose routines ignore the unit flag. Green CI here is not evidence about that either way, so if your probe-target work ends up exercising a setpoint path on a non-PBL board, that is worth more than any test we have.

trailro:
`2026.8.3` is out and carries everything this needed, so I have force-updated this PR to the split version at `a481e5f` rather than opening a new one — the review above is what shaped it and I would rather not orphan it.

Verified against the released package from PyPI, not an overlay of a branch: 97 tests, ruff, `ruff format --check` and mypy all green, on current `main` where you have already moved both pins.

The description is rewritten to describe what is actually here now. Squashed to a single commit, since the second one only undid the first and a two-commit history where half is reverted is not worth reviewing through.

Nothing outstanding from me. The remaining items in #321 that were waiting on this release — set 5's unit select and its two buttons, and set 7's reauth — are now unblocked too; I will send them in order rather than all at once, and the options-flow question on #277 still gates the two that need one.

dknowles2 (commented):
This is the split I asked for, and it is better than the version it replaces. The coordinator no longer knows the scratchpad key naming, the Fahrenheit convention, the wholesale-assign merge or the power-off wipe — all of that is behind `set_probe_target()` / `get_probe_targets()` now, and what is left reads like Home Assistant code.

**The three-way precedence survived the split intact, and the line you drew is the right one.** pytboss owns board-reported over scratchpad because both are things the grill told us; Home Assistant keeps the restored rung because it only exists as a consequence of HA outliving a store the grill wipes. Reading `pNTarget` from `self.data` directly rather than through the library, so an announced target shows on the push instead of at the next poll, is a real improvement over routing it through `get_probe_targets()`.

Volunteering the two bugs your own tests caught during the rewrite — the mutated return value and `native_value` rendering `165.0` — is worth more than the fixes themselves.

## What I verified against the released 2026.8.3

* `set_probe_target`, `get_probe_targets` and `probe_target_command` all present; `UnsupportedOperation` importable from `pytboss.exceptions`.
* `set_probe_target` raises `UnsupportedOperation` **only on the scratchpad path** when the grill is off — the board-command path sends regardless. Your `except UnsupportedOperation` handles that correctly in both directions.
* `get_probe_targets()` builds and returns a fresh dict, so your defensive copy is belt-and-braces rather than load-bearing. Worth keeping anyway; the comment states a convention that will still be true if the library changes.
* **Every one of the 137 models declares `meat_probes` of 2, 3 or 4** — none is 0 or `None`, so the `or 1` fallback at `number.py:48` never fires today.

**I could not run the tests.** They fail in my environment, but `tests/test_switch.py` fails identically on unmodified `main` — 9 failures either way — so it is environmental and CI is the real check. I am not implying I ran them.

## One edge case worth deciding on before this lands

Setup order in `__init__.py` is `async_config_entry_first_refresh()` at line 130, then `async_forward_entry_setups()` at line 143. So the coordinator's first refresh happens **before** any entity's `async_added_to_hass` populates `restored_targets`.

Most of the time that is fine, and I think deliberately so: if the grill is on at HA startup its store is intact, so `probe_targets` is authoritative and the restored value is not needed. If the grill is off, `_async_refresh_probe_targets` returns early leaving `_targets_seeded` false, and seeding runs properly when it comes on.

The gap is the third case: **grill power-cycles while HA is down, and is back on before HA starts.** Then the first refresh sees `moduleIsOn`, sets `_targets_seeded = True` against an empty `restored_targets`, and fetches a store that the power cycle wiped. Entities then restore their targets into `restored_targets` — where they are displayed via the third rung, but never written to the grill, because seeding only runs once per off-to-on transition and that transition already happened.

So the target shows in Home Assistant and is not in the grill's store, until the next power cycle. Given the store is a note rather than something the board acts on, the practical cost is small — but it is the one scenario the restore machinery exists for, and it is the one where it does not fire.

The cheap fix is to let the seeding gate be re-armed by the entities: have `async_added_to_hass` clear `_targets_seeded` when it restores a value the coordinator does not already have. Or seed on any refresh where `restored_targets` holds something `probe_targets` lacks, rather than only on the transition. Either is a few lines; I would rather you pick than have me guess at the shape.

Not blocking on it if you would rather land this and follow up — say which and I will merge.

## Two smaller things

`number.py:48` — `meat_probes or 1` produces one target entity for a grill declaring no probes. No model does, so it is unreachable, but `or 0` seems the safer direction for an unknown: no probes should mean no probe targets.

`number.py:110-114` — `available` now overrides the parent only to return it unchanged. The comment is worth keeping; the method is not. A comment on the class, or none at all, says the same thing without leaving a property that looks like it does something.

Behaviour changes 1–3 are all fine and all previously agreed. Labelled `enhancement`.

trailro:
All three done at `5a2dc2a`. The edge case was real and I would rather fix it than land it as a known gap.

**The seeding gap.** Your reading of the setup order is right: `async_config_entry_first_refresh()` runs before `async_forward_entry_setups()`, so a grill that power-cycled while Home Assistant was down is marked seeded against an empty `restored_targets` before any entity has restored anything. The target then shows in HA and never reaches the grill — the one case the restore machinery exists for.

I took your first option, re-arming, but put it behind a coordinator method rather than having entities write to a private flag:

```python
def note_restored_target(self, probe_number: int, temp: int) -> None:
    self.restored_targets[probe_number] = temp
    if probe_number not in self.probe_targets:
        self._targets_seeded = False
```

Chose that over "seed on any refresh where `restored_targets` holds something `probe_targets` lacks" because of what the two do when a target is removed elsewhere. Under the once-per-transition rule, a target the vendor's app clears stays cleared. Under the always-check rule, we would notice it missing and write ours back every poll — quietly overriding a deliberate change made on the phone. The re-arm keeps the transition semantics and only reopens the gate when a restore genuinely brings information the grill does not have.

Two tests, and I verified the first fails without the re-arm and passes with it:

- the power-cycle sequence you described end to end
- a restored target the grill *already* has is not rewritten, which is the regression the always-check version would have introduced

**`meat_probes or 1` → `or 0`.** Agreed, and I confirmed your count independently before changing behaviour on it: 94 models declare 2, 35 declare 3, 8 declare 4. None is 0 or `None`, so this is unreachable either way — but an unknown grill with no probes should get no probe targets.

**The `available` override is gone**, comment moved to the class docstring where it says the same thing without leaving a property that looks like it does something.

**On the things you checked:** that `set_probe_target` raises `UnsupportedOperation` only on the scratchpad path is the behaviour the `except` relies on in both directions, and I had not stated it explicitly — good to have it written down. And you are right that the defensive `dict()` copy is belt-and-braces against the current library; it stays because the comment states a convention rather than a fact about today's implementation.

Thank you for saying plainly that you could not run the tests and that `test_switch.py` fails identically on unmodified `main`. That is more useful than a green tick would have been.

99 tests, ruff, `ruff format --check` and mypy clean.



---

*Edited: I first wrote `de51c31` above. That SHA does not exist — the commit is `5a2dc2a`. Same mistake I made on pytboss#530 earlier today and apologised for: I composed the comment before running the push instead of taking the hash from its output. Fixing the habit rather than just the text — the comment gets written after the push from now on.*

dknowles2 (approved):
All three addressed, and the seeding fix is better than what I proposed.

**I suggested re-arming or always-checking and left the shape to you; you picked re-arming for a reason I had not thought of.** Always-checking would notice a target missing from the grill's store and write ours back every poll — so clearing a target from the vendor's app would silently undo itself. That is a worse failure than the one I was trying to fix, and it would have been very hard to diagnose. `note_restored_target()` keeps the once-per-transition semantics and only reopens the gate when a restore brings information the grill genuinely lacks.

Putting it behind a coordinator method rather than letting entities poke `_targets_seeded` is also the right call — the flag stays private and the invariant lives with the state it guards.

I traced both paths against the new code:

* **Power cycle while HA was down**: first refresh seeds against an empty set and sets the flag; `note_restored_target(2, 165)` finds `2 not in probe_targets` and clears it; next refresh seeds and writes. Matches `test_a_target_restored_after_a_power_cycle_still_reaches_the_grill`.
* **Target already on the grill**: `probe_targets = {2: 190}`, so the flag is left set, no write happens, and `probe_target(2)` still returns 190. That is the regression the always-check version would have introduced, and you wrote the test for it.

Neither can loop: `note_restored_target` is only reachable from `async_added_to_hass`, once per entity per start.

`or 0` and the removed `available` override are both done, and moving that comment to the class docstring says the same thing without leaving a property that looks like it does something.

**On the SHA correction you appended** — noted, and the part that matters is that you found it yourself and named the habit rather than the typo. Writing the comment after the push is the right fix.

**I could not run the tests**, same as before: `tests/test_number.py` fails on unmodified `main` here (5 failed, 1 passed), so it is environmental. CI is green — build, mypy, ruff, Hassfest, HACS — and that is the real check.

This is a good piece of work. It started as a change carrying a hundred lines of protocol knowledge in the wrong repo and ends as the Home Assistant half of a split that made the library better too. Merging.

PR #342, opened 2026-08-02
Keep switches available when the grill is off

trailro:
Set 8 of #321, and the last piece of it — the other half of the pair, dropping the unplugged-probe gate on target numbers, travelled in #341.

Switch availability is gated on `moduleIsOn`, so both switches disappear when the grill is off. This drops that gate.

**For the power switch the gate is self-referential rather than a policy choice.** Its `entity_description.key` *is* `moduleIsOn`, so it hides the entity on exactly the state the entity exists to report. It can never show `off` — it goes `unavailable` instead, and an automation waiting for the grill to go off never sees the transition, because `unavailable` is not a state you can trigger on the way you can trigger on `off`.

There is a test for that specifically: the switch reads `on`, then `off`, across a `moduleIsOn` change.

**The primer is the weaker half, and I would rather say so.** Priming a grill that is off does nothing, so `unavailable` is defensible there in a way it is not for the power switch. I have dropped it for both because that is what the roadmap item said and what I have been running, and because an entity that vanishes is a worse way to say "not now" than one that is present and off. **Happy to re-gate the primer alone** if you would rather — it is two lines and the power switch is the part that matters.

**What the gate protected is still protected.** `super().available` still gates on the connection, so a disconnected grill makes both switches unavailable exactly as before. There is a test for that too, since removing one gate and silently removing the other would be the easy mistake here.

`test_switch.py:53` encoded the old behaviour and now encodes the new — that is the assertion you flagged in the roadmap as being in the way, and it is updated rather than deleted.

Verified the new tests fail against the old gate rather than assuming they would: restoring it fails `test_the_power_switch_can_report_off` and `test_is_on_and_availability`, and nothing else.

92 tests, ruff, ruff format and mypy clean, on current `main` with `pytboss==2026.8.2`. No pytboss dependency — this is the only queued set that needs nothing from a release.

### DIFF
--- a/custom_components/pitboss/switch.py
+++ b/custom_components/pitboss/switch.py
@@ -47,9 +47,13 @@ class BaseSwitchEntity(BaseEntity, SwitchEntity):
 
     @property
     def available(self) -> bool:
-        """Return if the switch is available."""
-        if data := self.coordinator.data:
-            return bool(data.get("moduleIsOn", True)) and super().available
+        """Return if the switch is available.
+
+        Deliberately not gated on `moduleIsOn`. For the power switch that
+        gate was self-referential -- it hid the entity on exactly the state
+        the entity exists to report -- so it could never show `off`, and an
+        automation waiting for the grill to go off never saw it happen.
+        """
         return super().available
 
 
--- a/tests/test_switch.py
+++ b/tests/test_switch.py
@@ -50,14 +50,15 @@ async def test_is_on_and_availability(
 
     coordinator.async_set_updated_data({"moduleIsOn": False, "primeState": False})
     await hass.async_block_till_done()
-    # All switches are unavailable when the module (grill) is off, since
-    # availability is gated on the shared "moduleIsOn" key.
+    # Switches stay available and report `off` when the grill is off, so an
+    # automation can trigger on the transition rather than on the entity
+    # disappearing.
     power_state = hass.states.get("switch.mygrill_module_power")
     prime_state = hass.states.get("switch.mygrill_prime")
     assert power_state is not None
     assert prime_state is not None
-    assert prime_state.state == "unavailable"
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
The self-referential argument is the whole case and it is correct: `PowerSwitch`'s `entity_description.key` *is* `moduleIsOn`, so gating availability on it hid the entity on precisely the state the entity exists to report. It could only ever be `on` or `unavailable`, and `unavailable` is not something an automation can trigger a transition on the way `off` is.

**One thing in your favour that you did not claim.** You called the primer the weaker half and offered to re-gate it. I would leave it dropped, and not only because the roadmap said so: Home Assistant's `available` means the entity cannot be *read* — the device is unreachable — not that an action would be pointless. Priming a grill that is off is a no-op, but that is a fact about what the command does, not about whether the entity can be read. So the primer gate was a misuse of availability too, just a less damaging one than the power switch's. Dropping both is the more conventional shape, not just the one you have been running.

**The second test is the one I would have asked for.** Removing one gate and silently removing the other is exactly the easy mistake here, and `test_switches_still_go_unavailable_when_disconnected` pins it. `BaseEntity.available` still requires `api.is_connected()` and non-empty coordinator data, so a disconnected grill behaves as before.

Updating `test_switch.py:53` rather than deleting it is right too — it encoded the old behaviour deliberately, so it should encode the new behaviour deliberately.

**What I could not check.** I could not reproduce your negative verification. Every test in this repo fails on `bluetooth_adapters` in my environment, because Home Assistant's `bluetooth` component will not start there — so unlike the pytboss PRs today, I am taking "restoring the gate fails those two and nothing else" on your word rather than having run it. CI is green, which is the real check here.

No pytboss dependency, so this is the one queued item that needs nothing from a release. Labelled `enhancement`. Merging.

PR #346, opened 2026-08-03
Add a button platform: restart the controller, request fast updates

trailro:
Set 5 of #321, first of two. Needs `pytboss 2026.8.3`, which `main` already pins.

Adds a `button` platform with two entities, both wrapping RPCs that have public wrappers as of [pytboss#530](https://github.com/dknowles2/pytboss/pull/530).

**Restart controller** — `api.reboot()`. The standard remedy when the WiFi module wedges: it drops the connection briefly, the grill keeps running. Over four weeks on Bluetooth mine lost the grill repeatedly, once for 32 minutes, caused by the module rebooting itself; being able to do that deliberately is the point.

**Request fast updates** — `api.request_fast_updates()`, and this one is deliberately not created on every entry.

Despite the RPC's name (`PB.WiFiAwakeWDT`) it does not keep the module awake. It arms a five-minute countdown that makes the firmware reschedule its status-push timer to `wsFastInterval` instead of `wsSlowInterval` — 5 seconds instead of 60. The push it accelerates is `WS.send(wsConn, ...)`, so **it does nothing on Bluetooth or a local connection**, and nothing at all while the grill is off (`lastWasOn || moduleIsOn` guards the send).

So it is gated on `CONF_PROTOCOL == PROTOCOL_WSS`. I would rather not ship a button that is a no-op on the transport most people are using — and I say that as someone who ran exactly that button locally for weeks before reading the firmware carefully enough to notice.

There is a test for the gate, and I verified it fails without it: removing the condition makes `test_no_fast_updates_button_over_bluetooth` fail and nothing else. It follows the BLE setup pattern already in `test_init.py` rather than inventing a fixture for it.

**On `Sys.Reboot`, inherited from #530 and worth repeating here:** it is a Mongoose OS core RPC rather than one the grill's app registers, so its availability rests on that core service being enabled. `Sys.GetInfo` working on my grill is a positive signal but not proof for `Sys.Reboot` specifically, and I have not pressed this button on a grill mid-cook to find out. If it turns out to be absent on some board, the failure is an `RPCError` from a button press rather than anything structural.

95 tests, ruff, `ruff format --check` and mypy clean on current `main`.

The second half of set 5 — a select for the unit the grill itself works in — is next, and is independent of this.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -39,6 +39,7 @@ from .services import async_register_services
 
 PLATFORMS: list[Platform] = [
     Platform.BINARY_SENSOR,
+    Platform.BUTTON,
     Platform.CLIMATE,
     Platform.LIGHT,
     Platform.NUMBER,
--- a/custom_components/pitboss/button.py
+++ b/custom_components/pitboss/button.py
@@ -1,0 +1,81 @@
+"""Button platform for pitboss."""
+
+from __future__ import annotations
+
+from homeassistant.components.button import ButtonDeviceClass, ButtonEntity
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.const import CONF_PROTOCOL, EntityCategory
+from homeassistant.core import HomeAssistant
+from homeassistant.helpers.entity_platform import AddEntitiesCallback
+
+from .const import DEFAULT_PROTOCOL, DOMAIN, PROTOCOL_WSS
+from .coordinator import PitBossDataUpdateCoordinator
+from .entity import BaseEntity
+
+
+async def async_setup_entry(
+    hass: HomeAssistant, entry: ConfigEntry, async_add_entities: AddEntitiesCallback
+):
+    """Setup button platform."""
+    coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
+    assert entry.unique_id is not None
+    entities: list[ButtonEntity] = [
+        RestartControllerButton(coordinator, entry.unique_id)
+    ]
+    # The fast-updates request accelerates the grill's push to the Dansons
+    # relay and nothing else, so it would be a decorative button on any other
+    # protocol.
+    if entry.data.get(CONF_PROTOCOL, DEFAULT_PROTOCOL) == PROTOCOL_WSS:
+        entities.append(FastUpdatesButton(coordinator, entry.unique_id))
+    async_add_entities(entities)
+
+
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
Approving. The gate is the part that makes this worth having, and it is the right call.

**Not shipping a button that does nothing on the transport most people use** is exactly right, and the reasoning is now checkable rather than asserted — `wsFastInterval = 5` / `wsSlowInterval = 60` at `fake_firmware/init.js:38-39`, the tick rescheduling at `:466-473`, and `sendWSStatus()` guarded on `wsConn` and ending in a `WS.send` at `:176`. I checked those against the repo's firmware model when reviewing pytboss#530 and they hold. The `lastWasOn || moduleIsOn` guard at `:472` is the second condition, and you have it in the docstring.

Saying you ran that button yourself for weeks before reading the firmware closely enough to notice is the useful half of that paragraph. It is the reason the gate exists rather than a nicety.

## What I verified

* `reboot()` and `request_fast_updates()` both present in the installed `2026.8.3`, checked against the release rather than a checkout.
* `PROTOCOL_WSS` / `DEFAULT_PROTOCOL` resolve as used, and `DEFAULT_PROTOCOL` is `PROTOCOL_WSS`, so an entry predating the protocol option still gets the button.
* `unique_id` follows the existing `f"<name>_{entry_unique_id}"` convention used by every other platform here.
* `Platform.BUTTON` added to `PLATFORMS`.
* CI green: build, mypy, ruff, Hassfest and HACS.

## What I could not verify, and am not implying otherwise

**I could not run the tests.** They fail in my environment, but so does `tests/test_switch.py` on unmodified `main` — 9 failures there either way — so it is environmental and CI is the real check. That means I did not independently reproduce your claim that removing the gate makes `test_no_fast_updates_button_over_bluetooth` fail and nothing else. Reading it, the test looks like it would: it asserts the restart button is present and the fast-updates one is not, in the same setup, so it distinguishes "gate missing" from "platform broken". I am taking the mutation result on your word.

`Sys.Reboot` stays as flagged on #530 — a Mongoose core RPC absent from `fake_firmware` because the app never registers it, with `Sys.GetInfo` working as a positive signal rather than proof. Agreed that the failure mode is an `RPCError` from a press rather than anything structural, which is the right level of risk for a diagnostic button.

One thought, not a change request: the fast-updates button does nothing while the grill is off, and you have that in the docstring but not in `available`. Gating it on `moduleIsOn` would make the UI say what the docstring says. I can see the argument the other way — a button that vanishes is more confusing than one that is briefly pointless — so I would leave it to you.

PR #347, opened 2026-08-03
Add a select for the unit the grill itself works in

trailro:
Set 5 of #321, second of two. Independent of #346. Needs `pytboss 2026.8.3`, which `main` already pins.

Every board declares `set-celsius` and `set-fahrenheit`, but until [pytboss#520](https://github.com/dknowles2/pytboss/pull/520) neither was reachable without a private call, so `isFahrenheit` was the one grill setting Home Assistant could observe and not change. This adds a select for it, via the new `api.set_temperature_unit()`.

**This changes a setting on the grill**, not how Home Assistant presents values — every temperature the board reports follows it, and the panel changes too. That is deliberately a different thing from the display override in #277, and I have said so in the entity docstring so the two are not confused when both exist.

Only created when the board declares both commands.

**The interesting part is the pending state, and it is not cosmetic.** The board clears its cached status (`sc_11`/`sc_12`) the moment it forwards a command to the MCU, so the next poll still carries the old unit. Without holding the chosen value the UI snaps back to the previous option a second after the user picks one, then flips again when the MCU catches up.

So the entity shows what was asked for until the grill confirms it, then drops the pending value and follows the grill again. There is a test for the whole sequence — pick, a stale poll that must not undo it, then confirmation — and I verified it fails without the pending value rather than assuming it would.

The settle timer is tied to the entity with `async_on_remove`, so unloading the entry mid-wait cancels it instead of firing at a coordinator that has gone away.

**One thing I deliberately did not do:** no optimistic conversion of other entities' values. When the grill switches to Celsius every temperature it reports changes unit, and Home Assistant pins an entity's displayed unit at registration — so those entities convert into the pinned unit rather than adopting the new one. That is the same interaction I flagged on #341 and #277, and it belongs wherever the display-unit question is settled, not here.

96 tests, ruff, `ruff format --check` and mypy clean on current `main`.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -43,6 +43,7 @@ PLATFORMS: list[Platform] = [
     Platform.CLIMATE,
     Platform.LIGHT,
     Platform.NUMBER,
+    Platform.SELECT,
     Platform.SENSOR,
     Platform.SWITCH,
 ]
--- a/custom_components/pitboss/select.py
+++ b/custom_components/pitboss/select.py
@@ -1,0 +1,99 @@
+"""Select platform for pitboss."""
+
+from __future__ import annotations
+
+from homeassistant.components.select import SelectEntity
+from homeassistant.config_entries import ConfigEntry
+from homeassistant.const import UnitOfTemperature
+from homeassistant.core import HomeAssistant, callback
+from homeassistant.helpers.entity_platform import AddEntitiesCallback
+from homeassistant.helpers.event import async_call_later
+
+from .const import DOMAIN
+from .coordinator import PitBossDataUpdateCoordinator
+from .entity import BaseEntity
+
+# The MCU is polled every couple of seconds; give it time to report back.
+MCU_SETTLE_SECONDS = 6
+
+UNIT_CELSIUS = UnitOfTemperature.CELSIUS
+UNIT_FAHRENHEIT = UnitOfTemperature.FAHRENHEIT
+
+# Board command slugs that switch the unit the grill itself works in.
+UNIT_COMMANDS = {
+    UNIT_CELSIUS: "set-celsius",
+    UNIT_FAHRENHEIT: "set-fahrenheit",
+}
+
+
+async def async_setup_entry(
+    hass: HomeAssistant, entry: ConfigEntry, async_add_entities: AddEntitiesCallback
+):
+    """Setup select platform."""
+    coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
+    assert entry.unique_id is not None
+    commands = coordinator.api.spec.control_board.commands
+    if all(slug in commands for slug in UNIT_COMMANDS.values()):
+        async_add_entities([GrillTemperatureUnitSelect(coordinator, entry.unique_id)])
+
+
+class GrillTemperatureUnitSelect(BaseEntity, SelectEntity):
+    """The temperature unit the grill itself displays and reports in.
+
+    This changes a setting on the grill, not how Home Assistant presents
+    values -- every temperature the board reports follows it.
+    """
+
+    _attr_icon = "mdi:temperature-celsius"
+
+    def __init__(
+        self,
+        coordinator: PitBossDataUpdateCoordinator,
+        entry_unique_id: str,
+    ) -> None:
+        super().__init__(coordinator, entry_unique_id)
+        self._attr_unique_id = f"grill_temperature_unit_{entry_unique_id}"
+        self._attr_name = "Grill temperature unit"
+        # Set per instance rather than on the class: the base declares it as
+        # an instance variable, so a class-level list is both a mutable class
+        # attribute and a mypy override error.
+        self._attr_options = [UNIT_CELSIUS, UNIT_FAHRENHEIT]
+        self._pending_option: str | None = None
+
+    @property
+    def _reported_option(self) -> str | None:
+        if data := self.coordinator.data:
+            return UNIT_FAHRENHEIT if data.get("isFahrenheit") else UNIT_CELSIUS
+        return None
+
+    @property
+    def current_option(self) -> str | None:
+        # Show what we asked for until the grill confirms it: the board wipes
+        # its cached status the moment it forwards a command to the MCU, so
+        # the next poll still carries the old unit and the UI would appear to
+        # snap back.
+        return self._pending_option or self._reported_option
+
+    @callback
+    def _handle_coordinator_update(self) -> None:
+        if self._pending_option and self._reported_option == self._pending_option:
+            self._pending_option = None
+        super()._handle_coordinator_update()
+
+    async def async_select_option(self, option: str) -> None:
+        """Switch the unit the grill works in."""
+        await self.coordinator.api.set_temperature_unit(
+            fahrenheit=option == UNIT_FAHRENHEIT
+        )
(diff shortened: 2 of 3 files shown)


dknowles2 (commented):
The pending state is the right idea and the reasoning behind it is correct — the board wiping `sc_11`/`sc_12` the moment it forwards to the MCU means an immediate poll reads the old unit, and without holding the choice the UI would visibly snap back. Tying the settle timer to the entity with `async_on_remove` rather than the coordinator is the detail I would have expected to be missed.

Being explicit in the docstring that this changes the **grill's** setting rather than Home Assistant's presentation is worth as much as the code, given #277 exists and does the opposite thing under a similar name.

## Verified

* `set_temperature_unit(fahrenheit: bool)` present in the installed `2026.8.3`.
* Your "every board declares both" claim, executed rather than read: **all 17 reachable boards declare `set-celsius` and `set-fahrenheit`, so all 137 models get the select.** The gate is defensive rather than selective, which is fine — worth knowing it never excludes anyone today.
* **I could not run the tests.** `tests/test_number.py` fails on unmodified `main` here (5 failed, 1 passed), so it is environmental and CI is the real check. I did not reproduce your mutation result on the pending-state test.

## One thing I would fix before this lands

**`_pending_option` has no way out if the grill never confirms.**

It is cleared in exactly one place — `_handle_coordinator_update`, and only when `_reported_option` matches. `_async_confirm` triggers a refresh but never clears it. So if the MCU does not apply the change, `current_option` returns the pending value **for the life of the entity**, and no later poll can dislodge it. Home Assistant shows °C, the grill is in °F, and nothing reconciles them short of a restart.

The send failing is handled — the exception propagates before `_pending_option = option`, so a refused command leaves the state alone. The gap is the send succeeding and the setting not taking, which is exactly the case the settle window exists to cover.

One line, in `_async_confirm`:

```python
async def _async_confirm(self, _now) -> None:
    await self.coordinator.async_request_refresh()
    # One settle window is all the pending value is for. If the grill did
    # not take the change, follow the grill rather than keep asserting it.
    if self._pending_option:
        self._pending_option = None
        self.async_write_ha_state()
```

By the time the refresh returns, `_handle_coordinator_update` has already cleared it if the grill confirmed — so this only fires in the failure case, and turns a permanent wrong reading into a six-second one. This is the same class of thing as the seeding gap on #341: a value shown in Home Assistant that the grill does not have and never will.

## One smaller thing

`async_on_remove` appends to the entity's removal list and is never trimmed, so every selection leaves another timer-cancel behind for the entity's lifetime. Harmless in effect — a fired timer's cancel is a no-op — but it grows without bound, and two selections inside six seconds queue two refreshes. Holding a single handle and cancelling the previous one is the usual shape:

```python
if self._cancel_settle is not None:
    self._cancel_settle()
self._cancel_settle = async_call_later(...)
self.async_on_remove(self._cancel_settle)
```

Agreed on what you left out. No optimistic conversion of other entities is right, and the registry-pinning interaction genuinely belongs wherever the display-unit question gets settled rather than here — I have made the same argument on #277, where I think it is decisive against that PR.

Labelled `enhancement`. Happy to merge once the pending-state escape is in; say if you would rather land it and follow up.

trailro:
Both fixed at `02ae60b`.

**The pending value now has a way out.** You are right that the failure mode is the bad one: the send succeeding and the setting not taking leaves the entity asserting a unit the grill will never report, permanently. Took your `_async_confirm` change, with the reasoning in a comment so the next reader knows why it only fires in the failure case.

There is a test for it — select, let the window pass with the grill still reporting the old unit, entity follows the grill again — and I verified it fails without the escape and passes with it.

**On the timer, I went slightly further than your snippet.** Yours still calls `async_on_remove` on every selection, so the removal list keeps growing even though only one timer is live at a time. Registering the cancel once and letting it cancel whatever handle is current keeps both properties:

```python
async def async_added_to_hass(self) -> None:
    await super().async_added_to_hass()
    self.async_on_remove(self._cancel_pending_settle)

@callback
def _cancel_pending_settle(self) -> None:
    if self._cancel_settle is not None:
        self._cancel_settle()
        self._cancel_settle = None
```

**And the test I first wrote for that was worthless, which is worth saying.** I asserted that three selections inside the window produce one `get_state()`. It passed with the fix and passed just as well without it — the coordinator's debouncer collapses the refreshes regardless, so I was testing the debouncer.

It now asserts the two things that are actually mine: the entity's removal list does not grow across selections, and `_async_confirm` fires once rather than once per selection. Against the old pattern that fails with `assert 8 == 5`, so it has teeth now.

I would not have caught that without going back to check the mutation, which is the habit your reviews have pushed me into.

**On the boards:** thank you for executing the "every board declares both" claim rather than reading it. All 17 reachable boards, all 137 models — so the gate never excludes anyone today and is defensive rather than selective. Worth having in the thread.

And noted on the tests failing in your environment. `tests/test_number.py` failing on unmodified `main` there matches what you saw on #341; CI is the real check.

108 tests, ruff, `ruff format --check` and mypy clean, rebased on current `main` — `main` moved twice while I was working on this, first #346 then #341, which is why it went BEHIND twice.

dknowles2 (approved):
Both fixed, and the timer handling is better than what I suggested.

You are right that my snippet still called `async_on_remove` on every selection — it fixed the multiple-live-timers half and left the growing-list half exactly where it was. Registering `_cancel_pending_settle` once in `async_added_to_hass` and letting it cancel whichever handle is current gets both, and it reads better besides. That is the third time tonight you have taken an option I offered and returned something tighter than the option.

I traced the new lifecycle: `_cancel_settle` holds at most one handle; `async_select_option` cancels before re-arming; `_async_confirm` clears the handle first because the timer has already fired; removal cancels whatever is live. No path leaves a timer behind or a stale handle set.

**The escape does what it needs to.** Clearing the pending value after the refresh means `_handle_coordinator_update` has already dealt with the success case, so the clear only bites when the grill did not take the change — turning a permanent wrong reading into a six-second one. The comment says exactly that, which is the part that stops someone "simplifying" it later.

## The part of your comment that matters most

**Reporting that your first timer test was worthless is worth more than the fix.** Asserting three selections produce one `get_state()` passed with and without the change because the coordinator's debouncer collapses them regardless — so it was testing Home Assistant, not this entity. Catching that requires running the mutation *expecting* it to fail and being suspicious when it does not, which is the opposite of how a passing test usually feels.

The replacement asserts the two things that are actually yours — the removal list not growing, and `_async_confirm` firing once — and `assert 8 == 5` against the old pattern is a real failure rather than a coincidence.

I could not reproduce that myself: `tests/test_number.py` fails on unmodified `main` here, so the suite is environmental for me and CI is the real check. I am taking the mutation results on your word, as before, and saying so rather than implying otherwise.

## Verified

* `CALLBACK_TYPE` imported from `homeassistant.core`, so the annotation resolves.
* `set_temperature_unit(fahrenheit: bool)` present in the installed `2026.8.3`.
* Merge base is current `main` (`710bf25`), and the change is genuinely three files — I checked, because my first diff of this branch was against a stale local `main` and wrongly showed #341's coordinator work as part of it. That was my error, not a rebase problem.

Agreed on what you left out — no optimistic conversion of other entities, and the registry-pinning interaction belonging wherever the display-unit question gets settled. Merging.

That closes set 5 and set 6. Worth saying plainly: across the last several of these you have found the better answer to my own suggestion more often than not, and the two times you were wrong you went and checked the primary source rather than arguing from the diff.

Issue #344, opened 2026-08-02
Local WiFi transport, adaptive polling and reauth (set 7 of #321)


trailro:
Splitting set 7 of #321 out, because what we learned on [pytboss#505](https://github.com/dknowles2/pytboss/issues/505) changes its shape enough that tracking it as one line on a roadmap is no longer honest.

Three items, listed as 16–18 there: **local WiFi transport**, **adaptive polling**, **reauth flow**.

## The finding that changes it

The local RPC interface is **not available on every grill, and there is no property that predicts it.**

| grill | board | firmware | `http.enable` |
|---|---|---|---|
| PB1600PS1 (mine) | PBL | 0.5.7 | `true` |
| @Giordano22's | PBL | — | `true` |
| PBV4PS2 (@dknowles2's) | PBL | 0.5.7 | **`false`** |

Same board, same firmware version, opposite setting. And it is not something the firmware decides: none of the nine mJS releases (0.2.3 → 0.6.0) touches `http.enable` — the only config the app reads is `app.wsFastInterval`, `app.wsSlowInterval` and `device.id`. No `conf0.json`/`mos.yml` ships alongside the JS either. It lives in the device's own Mongoose OS config.

So a gate on model, board or firmware version would each be wrong. **The only correct gate is asking the grill**, and that needs no new API:

```python
http = await api.config.get_config("http")
if http.get("enable"): ...
```

`config.get_config()` is public in pytboss today, and the integration is already connected over Bluetooth before it would offer this.

## The review problem, stated plainly

**@dknowles2 cannot exercise this feature on his own hardware.** His grill has the interface off, and the test suite is fully mocked with `pytest-socket` blocking real connections, so no test in this repo can reach a grill either — in either direction. My passing tests would prove nothing about the transport, and I would rather say that up front than let green CI imply otherwise.

Two things follow, and I think they make it reviewable anyway:

- Evidence should be independently checkable rather than "it works here" — the firmware is on a public CDN (`https://firmware.dansonscorp.com/<version>/init.js`, no auth) so any claim I make about it can be verified.
- **His grill is the better test case for the half that matters more.** He cannot test that the transport works, but he can test exactly what happens on a grill that does not have it: that the option is not offered, that the reason is legible, and that nothing breaks. That path will reach far more users than the happy one.

### 16 — Local WiFi transport

A `Transport` subclass speaking JSON-RPC over `ws://<host>/rpc`, plus a third protocol option in the config flow, gated on the probe above.

[...]

PR #348, opened 2026-08-03
Add a target-reached sensor for every probe

trailro:
Set 6 of #321, last of three. Builds on #341, which landed the targets this reads.

The grill acts on the control probe only, so it raises nothing for the rest — and #341 gave every probe a target it can hold, from three sources: one the board reports, one in its virtual data store, or one restored here across a restart. This turns those into a signal to automate on.

One binary sensor per probe, up to `meat_probes`, comparing the reported temperature against `coordinator.probe_target()` — so it works for probes the grill knows nothing about, which is the point.

**Deliberately `off` rather than unknown when there is no probe or no target.** An automation triggering on off-to-on does not fire reliably out of `unknown` or `unavailable`, and this entity exists to be triggered on. Reporting "I don't know" would be more honest and less useful; the state means "not reached", which is true in both cases.

That is also why there is no `available` override here. On the earlier version of #341 I had one that returned the parent unchanged to hang a comment on, and you were right that a property which looks like it does something is worse than a comment on the class. The reasoning is in the class docstring instead.

Four tests:

- one sensor per probe, and no third on a two-probe grill
- follows a board-reported target across the threshold
- **`off` rather than unknown** for an unplugged probe, and for a probe with no target
- **a target the grill does not report at all** — set only in `restored_targets`, crossing 165 from 164 to 166

I verified the last one has teeth by reading the target straight from `data["pNTarget"]` instead of through `probe_target()`: it fails, and nothing else does. That is the test that distinguishes this from a sensor that only works on the control probe.

112 tests, ruff, `ruff format --check` and mypy clean on current `main`.

With this, sets 1, 2, 3, 4, 5, 6 and 8 from #321 are complete. Set 7 is the local transport, adaptive polling and reauth, tracked separately in #344 — and the transport half of it is now largely your #543.

### DIFF
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -123,6 +123,10 @@ async def async_setup_entry(
     for entity_description in ENTITY_DESCRIPTIONS:
         entities.append(BinarySensor(coordinator, entry.unique_id, entity_description))
     entities.append(ConnectivitySensor(coordinator, entry.unique_id))
+    for probe_number in range(1, (coordinator.api.spec.meat_probes or 0) + 1):
+        entities.append(
+            ProbeTargetReachedSensor(coordinator, entry.unique_id, probe_number)
+        )
     async_add_entities(entities)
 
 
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Small, and it only works because of the split it sits on top of. Comparing against `coordinator.probe_target()` rather than `data["pNTarget"]` is what makes this a signal for every probe instead of just the control one, and the test that pins it — a target present only in `restored_targets`, crossing 165 from 164 to 166 — is the one that would catch a future "simplification" back to the state key. Verifying that by mutation rather than assuming it is the right instinct.

**Off rather than unknown is the right call and the reasoning is the useful part.** An entity that exists to be triggered on should not spend its idle life in a state that automations fire unreliably out of. "Not reached" is true whether the probe is absent or the target is unset, so the state is honest as well as useful — those two do not always coincide and it is worth that you said which you were optimising for.

And noted on dropping the `available` override rather than carrying one to hang a comment on. That is the second time you have applied a point from one review to a place I had not raised it.

## Verified

* Merge base is current `main` (`ef752f5`) and the change really is two files — I pulled first, having previously mistaken a stale local `main` for a branch problem.
* **Units line up.** `pNTemp` and `probe_target()` are both in the grill's own unit — `probe_target`'s docstring says so and #341's precedence chain converts the virtual-data rung through pytboss — so the comparison never mixes scales. That is the thing most likely to be quietly wrong in a sensor like this, and it is right.
* `meat_probes or 0`, consistent with the fix on #341, so an unknown grill gets no probe sensors rather than one.
* `probe_label` already imported and used by the existing error sensors, so naming stays consistent with the rest of the platform.
* **I could not run the tests.** `tests/test_binary_sensor.py` fails 5/5 on unmodified `main` here, so it is environmental as before and CI is the real check. I did not reproduce your mutation on the `restored_targets` test.

## One observation, not a change request

`temperature >= target` will flap if a reading sits on the threshold. Probe readings are whole degrees and meat near its target rises slowly, so a degree of noise can cross back and forth several times, and an automation on `to: "on"` fires on each crossing — three notifications instead of one.

A user can fix that themselves with `for: "00:01:00"` on the trigger, which is the ordinary Home Assistant answer and probably where it belongs. Hysteresis in the entity would be the alternative, and I do not think it is worth the state it would need. Recording it so that if someone reports repeat notifications, the cause is already written down rather than rediscovered.

That is sets 1 through 6 and 8 from #321 done. Set 7's transport half is pytboss#543 — which, thanks to your #546, now has both of my bugs out of it and a real grill's `get_state()` behind it rather than a test server I wrote from the same assumptions as the code.

PR #349, opened 2026-08-03
Expose whether the grill has a meat probe control port

trailro:
The last unshipped half of item 12 in #321. That item was "label probe ports MPC / MP1 / MP2 … **plus a capability sensor**"; #333 landed the labels and I never sent the sensor. Going through the roadmap to close it out is what surfaced it.

A binary sensor for `Grill.has_mpc` — whether the grill has a meat probe control port, and therefore whether probe 1 is the one the board acts on.

**It is partly redundant and I want to be upfront about that.** #333 already says it to anyone reading a dashboard: the entity is called MPC. What a name cannot do is be read by an automation, and "does this grill act on a probe at all" is the question behind most of what the probe entities are for — whether a target will be enforced or is only a note.

If you think the naming carries it, say so and I will drop this; it is the one roadmap item where I would not argue hard.

Always available, like `ConnectivitySensor`, because it answers a question about the model rather than about the connection — a grill being unreachable does not make it stop having the port. There is a test for that.

Three tests: on for a grill with the port, off for one without, and available while disconnected. The second swaps the spec with `dataclasses.replace` rather than parametrizing the model, because `test_binary_sensor.py` sets `model` at module level and a second `parametrize` on the same argument is a collection error.

115 tests, ruff, `ruff format --check` and mypy clean on current `main`.

**With this, every item in #321 that belongs in this repo is shipped.** 1–15 and 19 are in `main`; 16–18 are set 7 and live in #344, where the transport half is now your #543.

### DIFF
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -123,6 +123,7 @@ async def async_setup_entry(
     for entity_description in ENTITY_DESCRIPTIONS:
         entities.append(BinarySensor(coordinator, entry.unique_id, entity_description))
     entities.append(ConnectivitySensor(coordinator, entry.unique_id))
+    entities.append(MeatProbeControlSensor(coordinator, entry.unique_id))
     for probe_number in range(1, (coordinator.api.spec.meat_probes or 0) + 1):
         entities.append(
             ProbeTargetReachedSensor(coordinator, entry.unique_id, probe_number)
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
You asked directly whether the naming carries it, so here is a real answer rather than a wave-through: **keep it**, with one caveat I want on the record so David can overrule us both.

**For it:** an automation cannot read an entity's name, and `has_mpc` is not decoration — it decides whether a probe target is a value the board enforces or a note in a scratchpad. Everything else in the probe entities behaves differently depending on the answer, so making it machine-readable is worth one diagnostic entity. It is also the half of item 12 that was specified and never shipped, rather than something new.

**Against, and it is not nothing:** Home Assistant's quality-scale guidance leans away from entities whose state can never change, and this one is fixed for the life of the device. The idiomatic alternative is an attribute on an entity that does change, read with `state_attr()`. I do not think that is better here — it buries a capability inside an unrelated entity's payload — but it is the argument someone will eventually make, and it is a fair one.

I checked the split it is describing: **33 of 137 models have the port, 104 do not**, so it is a real distinction rather than a constant dressed up as a sensor.

## Verified

* Merge base is current `main` (`fc7eadb`), two files, no stale-checkout confusion this time.
* `available` returning `True` matches `ConnectivitySensor` exactly, and for the same shape of reason — that entity says the grill is unreachable, this one answers a question about the model. Neither depends on the connection.
* `has_mpc` is a plain `bool` field on `Grill`, not derived, so there is nothing to go stale.
* `coordinator.has_mpc` already exists and is what `probe_label` uses, so the sensor and the port naming cannot disagree.
* **I could not run the tests.** `tests/test_binary_sensor.py` fails 9/9 on unmodified `main` here, so it is environmental as before and CI is the real check.

The `dataclasses.replace` workaround is the right shape — a second `parametrize` on `model` against a module-level one is a collection error, and swapping the spec keeps the test about `has_mpc` rather than about a particular grill.

## One cosmetic thing

`mdi:thermometer-check` is now on two binary sensors on the same device — this one and `ProbeTargetReachedSensor` from #348. On a dashboard "Meat probe control" and "MPC target reached" will show the identical icon, which is the one place a person is most likely to confuse a capability with a state. Something like `mdi:thermometer-probe` here would separate them. Not worth a round trip on its own; fold it in if you touch this file again.

That is every item in #321 that belongs in this repo. For what it is worth, the roadmap ends in a better state than it started: the two changes I pushed back hardest on — the probe targets and the unit select — are both materially different from what was first proposed, and better for it.

PR #350, opened 2026-08-03
Ask for the password when the grill rejects it

trailro:
Item 18 of #321, tracked in #344. **Independent of the options-flow question** — reauth is a config flow step, not an options one, so it does not touch #277 or wait on it. The other two items in #344 are the local transport (now your pytboss#543) and adaptive polling (which does want the flow).

`PB.GetState` is password-gated. A stale password therefore does not just break writes — it takes the whole integration down, and today it retries forever without ever saying why.

This raises `ConfigEntryAuthFailed` on a rejected password, so Home Assistant asks for a new one instead:

```python
except Unauthorized as ex:
    raise ConfigEntryAuthFailed("The grill rejected the stored password") from ex
except RPCError as ex:
    raise UpdateFailed(str(ex)) from ex
```

**Clause order matters here and is deliberate.** Since [pytboss#524](https://github.com/dknowles2/pytboss/pull/524), `Unauthorized` subclasses `RPCError` — that was the point, so existing `except RPCError` handlers kept working. It also means catching `RPCError` first would swallow the 401 silently, which is the bug this replaces rather than a new one.

That PR is what made this possible at all. Before it the only way to tell a 401 from any other RPC failure was matching on the message text, which I was doing downstream and would not have proposed here.

**The password field is optional and defaults to empty**, because clearing the grill's password is a supported thing to do — the firmware's `checkPassword` returns true immediately when the stored password is empty, so an empty entry is "no authentication" rather than a blank guess.

Four tests:

- a 401 during setup puts the entry in `SETUP_ERROR` with a `reauth` flow in progress
- **any other `RPCError` does not** — it stays `SETUP_RETRY` with no flow, which is the half that would regress silently if the clause order were ever flipped
- reauth writes the new password to the entry and reloads
- an empty password is accepted

Verified the first fails without the change rather than assuming it would.

119 tests, ruff, `ruff format --check` and mypy clean on current `main`.

### DIFF
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from typing import Any
 
 import voluptuous as vol
@@ -106,6 +107,35 @@ class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
                 }
             ),
             description_placeholders={"name": self._device_id},
+        )
+
+    async def async_step_reauth(
+        self, entry_data: Mapping[str, Any]
+    ) -> ConfigFlowResult:
+        """Handle a password the grill no longer accepts."""
+        return await self.async_step_reauth_confirm()
+
+    async def async_step_reauth_confirm(
+        self, user_input: dict[str, Any] | None = None
+    ) -> ConfigFlowResult:
+        """Ask for the grill's current password.
+
+        Optional and defaulting to empty, because clearing the password on the
+        grill is a supported thing to do -- the firmware treats an empty one
+        as no authentication at all.
+        """
+        reauth_entry = self._get_reauth_entry()
+        if user_input is not None:
+            return self.async_update_reload_and_abort(
+                reauth_entry,
+                data_updates={CONF_PASSWORD: user_input.get(CONF_PASSWORD, "")},
+            )
+        return self.async_show_form(
+            step_id="reauth_confirm",
+            data_schema=vol.Schema({vol.Optional(CONF_PASSWORD, default=""): str}),
+            description_placeholders={
+                "name": reauth_entry.data.get(CONF_DEVICE_ID, "")
+            },
         )
 
     async def async_step_reconfigure(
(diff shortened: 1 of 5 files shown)


trailro:
Cross-note with [#351](https://github.com/dknowles2/ha-pitboss/pull/351): the two conflict, but only textually — both append tests to the end of `tests/test_init.py` and both touch its import block. One hunk, no shared logic. Whichever lands second I will rebase; no action needed on your side.

They are independent otherwise: this one does not read the other's code and neither changes behaviour the other tests.

dknowles2 (approved):
**The clause order is the whole PR and you are right that getting it backwards would be silent.** I confirmed the hierarchy rather than taking it from the description:

```
Unauthorized MRO: ['Unauthorized', 'RPCError', 'Error', 'Exception']
issubclass(Unauthorized, RPCError): True
```

So `except RPCError` first would catch the 401 and turn a permanent auth failure into an endless retry — which, as you say, is the behaviour being replaced rather than something new. Making `Unauthorized` a subclass on pytboss#524 was deliberately compatible, and this is the first place that compatibility has a sharp edge. The test asserting a *non*-401 `RPCError` stays `SETUP_RETRY` with no flow is the one that keeps the order honest, and it is the right test to have written.

I also checked nothing upstream swallows it first. `ping()` is `RPC.Ping`, unauthenticated; `_async_refresh_sys_info` catches broadly but runs before `get_state()` and cannot mask it; and `_async_refresh_probe_targets` — which does catch `Exception` — is only reached after `get_state()` has already returned, so a rejected password raises from `get_state()` and lands on your clause. No path around it.

## Verified

* Merge base is current `main` (`d4f88fd`).
* `_get_reauth_entry` and `async_update_reload_and_abort` both exist in the pinned HA 2026.7.4, so this is not written against a newer API than the manifest allows.
* `reauth_confirm` and `reauth_successful` are both in `translations/en.json`, so the form and the abort have strings rather than raw keys.
* **I could not run the tests.** `tests/test_config_flow.py` and `tests/test_init.py` fail 19/19 on unmodified `main` here, so it is environmental as always and CI is the real check. I did not reproduce your mutation.

**The optional, empty-defaulting password is the right call and worth having argued for.** An empty entry meaning "no authentication" rather than "a blank guess" is a fact about `checkPassword` returning true immediately on an empty stored password, not a UI convenience — and a reauth form that forces a non-empty value would make clearing the grill's password an unrecoverable state. That is exactly the kind of thing that only bites someone months later.

Good that this turned out not to need the options flow. It means #344 is no longer entangled with #277, which I have recommended closing on separate grounds.

PR #351, opened 2026-08-03
Poll faster while the grill is running

trailro:
Item 17 of #344, and **without the options flow** — which removes its dependency on #277 entirely.

Polling is currently a flat 30 s. Nothing interesting changes on a grill sitting in standby and each cycle costs several round trips, but once it is lit the temperatures move and the delay is what the user actually feels. So: **10 s while `moduleIsOn`, 60 s otherwise.**

**This makes fewer network requests, not more**, which is the part I would have got wrong by guessing:

| | now (flat 30 s) | this |
|---|---|---|
| a day with no cooking | 2880 | 1440 |
| a day with a four-hour cook | 2880 | 2640 |

A grill is idle almost all the time, so backing off in standby more than pays for polling hard while it runs. 60 s rather than the 30 s I originally planned for standby is what turns it from roughly neutral into a reduction.

**Constants rather than options.** I had this written against an options flow and held it behind #277; making the two intervals module constants gets the behaviour in now and leaves them as exactly the two fields that would become options later. Nothing here has to change when the flow arrives — the values move, the logic does not.

**Applied on the push path too, not only on poll.** On Bluetooth a power change arrives through `_on_state_update`, and the interval should follow it rather than wait for the next poll to notice.

Two tests, both verified to fail without the change:

- the interval follows the grill on and off
- **a partial frame does not read as "off"** — the board answers with two frames and clears them independently, so a frame carrying no `moduleIsOn` must not back the polling off. The interval is decided from the merged state, not the frame in isolation. Applying it to the raw frame instead makes that test fail and nothing else, which is what makes it worth having.

That second one is the same hazard as #326: a frame that is missing a key is not a frame that says the key is false.

117 tests, ruff, `ruff format --check` and mypy clean on current `main`.

With this, #344 is down to item 16 — the local transport, which is your [pytboss#543](https://github.com/dknowles2/pytboss/pull/543) and waits on a release.

### DIFF
--- a/custom_components/pitboss/const.py
+++ b/custom_components/pitboss/const.py
@@ -8,7 +8,18 @@ LOGGER: Logger = getLogger(__package__)
 NAME = "PitBoss"
 DOMAIN = "pitboss"
 MANUFACTURER = NAME
-PING_INTERVAL = timedelta(seconds=30)
+# How often the coordinator polls, depending on what the grill is doing.
+#
+# Nothing interesting changes on a grill sitting in standby, and each cycle
+# costs several round trips -- but once it is lit, temperatures move and the
+# delay is what the user actually feels. Against the flat 30s these replace,
+# an idle day halves its requests and even a four-hour cook makes slightly
+# fewer of them overall.
+#
+# Constants rather than options because this integration has no options flow
+# yet. They are the two values that would become fields when it grows one.
+ACTIVE_SCAN_INTERVAL = timedelta(seconds=10)
+STANDBY_SCAN_INTERVAL = timedelta(seconds=60)
 # Diagnostics change slowly; no need to read them at the poll rate.
 SYS_INFO_INTERVAL = 60.0
 PROTOCOL_WSS = "wss"
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -19,7 +19,13 @@ from pytboss.exceptions import (
 )
 from pytboss.grills import StateDict
 
-from .const import DOMAIN, LOGGER, PING_INTERVAL, SYS_INFO_INTERVAL
+from .const import (
+    ACTIVE_SCAN_INTERVAL,
+    DOMAIN,
+    LOGGER,
+    STANDBY_SCAN_INTERVAL,
+    SYS_INFO_INTERVAL,
+)
 
 
 class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
@@ -38,7 +44,11 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
     ) -> None:
         """Initialize the coordinator."""
         super().__init__(
-            hass=hass, logger=LOGGER, name=DOMAIN, update_interval=PING_INTERVAL
+            hass=hass,
+            logger=LOGGER,
+            name=DOMAIN,
+            # Starts in standby; the first state that arrives corrects it.
+            update_interval=STANDBY_SCAN_INTERVAL,
         )
         self.device_info = device_info
         self.api = api
@@ -186,9 +196,23 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         merged.update(state)
         return merged
 
+    def _apply_poll_interval(self, state: StateDict) -> None:
+        """Poll hard while the grill runs, back off in standby."""
+        wanted = (
+            ACTIVE_SCAN_INTERVAL if state.get("moduleIsOn") else STANDBY_SCAN_INTERVAL
+        )
+        if self.update_interval != wanted:
+            self.logger.debug("Polling every %s", wanted)
+            self.update_interval = wanted
+
     async def _on_state_update(self, data: StateDict) -> None:
         self.logger.debug("Received data: %s", data)
-        self.async_set_updated_data(self._merge_state(data))
+        merged = self._merge_state(data)
+        # Applied on the push path too: on Bluetooth a power change arrives
+        # this way, and the poll interval should follow it without waiting
+        # for the next poll to notice.
+        self._apply_poll_interval(merged)
+        self.async_set_updated_data(merged)
 
     async def _start_api(self) -> None:
         try:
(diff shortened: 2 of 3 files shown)


trailro:
Cross-note with [#350](https://github.com/dknowles2/ha-pitboss/pull/350): the two conflict, but only textually — both append tests to the end of `tests/test_init.py` and both touch its import block. One hunk, no shared logic. Whichever lands second I will rebase; no action needed on your side.

They are independent otherwise: this one does not read the other's code and neither changes behaviour the other tests.

dknowles2 (approved):
**I re-derived all three numbers rather than reading the table, and they are exact:**

```
flat 30s, per day : 2880
new, idle day     : 1440
new, 4h cook      : 2640   (4h at 10s = 1440, 20h at 60s = 1200)
```

So the claim holds: this reduces polling on every realistic day, including one with a long cook, and the 60 s standby is what does the work. Saying you would have got that wrong by guessing is the useful part — "poll faster" reads like more traffic and it is not.

**Deciding the interval from the merged state rather than the frame is the load-bearing detail.** The board answers in two frames and clears them independently, so a frame with no `moduleIsOn` is not a frame saying the grill is off — and reading it that way would back polling off to 60 s in the middle of a cook, which is exactly the moment the change exists to improve. Same hazard as #326, and worth that you made it a test rather than a comment.

Applying it on the push path too is right: on Bluetooth a power change arrives through `_on_state_update`, and waiting up to 60 s to notice the grill lit would undo most of the benefit on the transport where it matters most.

## Verified

* Merge base was current `main`, and `PING_INTERVAL` is gone from the tree — no dangling import or stale constant left behind.
* Setting `self.update_interval` reschedules through the coordinator's own setter, so the new cadence takes effect without a restart.
* Starting in standby and letting the first state correct it is the safe default — a grill that turns out to be lit costs one 60 s cycle, where the reverse would poll a sleeping grill hard until the first frame arrived.
* **I could not run the tests.** `tests/test_init.py` fails on unmodified `main` here (19 failures across it and `test_config_flow.py`), so it is environmental and CI is the real check. I did not reproduce your two mutations.

## One question, not a change request

The daily totals go down, but the **sustained rate during a cook goes up threefold** — and that is the number BLE feels. Per cycle while the grill is on this issues `RPC.Ping`, `PB.GetState` and `PB.GetVirtualData` (via `get_probe_targets`), with `Sys.GetInfo` held off by `SYS_INFO_INTERVAL`. That is roughly 18 round trips a minute at 10 s against 6 today.

On the websocket that is nothing. On Bluetooth, where you have had the grill drop for half an hour at a time, three times the sustained traffic during exactly the hours the connection is under load is the kind of thing that only shows up on hardware. You run this on BLE in production — has the 10 s cadence been through a real cook there? If it has, that is worth a line in the PR, because it is the one risk the arithmetic does not cover. If it has not, I would still take this, but it is the first thing I would suspect if drop reports follow.

Worth noting `get_probe_targets` only costs a round trip while the grill is on, which is also the only time we poll at 10 s — so the two multiply rather than one offsetting the other.

Labelled `enhancement`. Merging; the question stands on its own and does not need to hold the change.

That leaves #344 at item 16 alone — the local transport, which is pytboss#543 and waits on David.

dknowles2:
Approved above, but **this now conflicts with `main`** — #350 landed first and the two of you appended tests at the same point in `tests/test_init.py`. Nothing in `coordinator.py` collides; the imports and the `_async_update_data` clauses merged cleanly. It is purely both-added, and the resolution is to keep both blocks:

```
unmerged: tests/test_init.py
<<<<<<< HEAD    test_the_poll_interval_follows_the_grill / test_a_partial_frame_does_not_change_the_interval
=======
>>>>>>> main    test_a_rejected_password_asks_for_a_new_one / test_other_rpc_errors_do_not_ask_for_a_password
```

A rebase on `a50f98f` is all it needs; all four tests survive unchanged and `ruff format` wants one blank-line fix afterwards.

**I need to own something here.** I resolved it locally to check it really was trivial, and then pushed to the wrong remote — so there is now a stray `feat/adaptive-polling` branch on `dknowles2/ha-pitboss` at `7237cea` that should not exist. It is not attached to this PR, does not affect `main`, and this PR's head is untouched at `104da20a`. I cannot delete it from here, so I have flagged it for David.

I should not have gone as far as pushing at all. The conflict is yours to resolve on your own branch and you have been turning these around in minutes; me reaching into it saved nothing and cost a stray ref. Rebase whenever suits and I will merge.

The question in my review stands separately and does not block: has the 10 s cadence been through a real cook on Bluetooth? It roughly triples sustained round trips during exactly the hours the BLE link is under load.

trailro:
Rebased onto `a50f98f` at `f43c250` — #350 landed and the two conflicted textually where I said they would, one hunk at the end of `tests/test_init.py`. 121 tests, ruff, format and mypy clean.

**On the BLE question: no, it has not, and I should be plain about that rather than let the approval carry it.** My grill runs on the local protocol, not Bluetooth — `protocol: local`, straight to the grill's own `/rpc`. So the 10 s cadence has been through real cooks, but never over BLE, which is the transport your question is about and the one where I have the drop history.

Your arithmetic matches mine: three round trips a cycle while the grill is on — `RPC.Ping`, `PB.GetState`, and `PB.GetVirtualData` through `get_probe_targets` — so 18 a minute at 10 s against 6 today. And you are right that the two multiply rather than cancel: `get_probe_targets` only costs a round trip while the grill is on, which is exactly when we poll hardest.

So the honest position is that the risk you name is real, unmeasured, and lands on the transport least able to absorb it. Two things that bound it, neither of which is evidence:

- The 32-minute drop I have cited was the ESP module rebooting itself while polling every 30 s, so at least that failure was not load-induced.
- `PB.GetVirtualData` is skipped entirely when the grill is off, so standby is genuinely one ping plus one state read per minute.

If drop reports do follow, the first thing I would try is dropping the active interval to 15 s rather than reverting — it keeps most of the responsiveness and cuts the sustained rate by a third. Happy to put that in a comment next to the constant so whoever hits it does not have to rediscover the reasoning.

I will report back once I have run a cook over Bluetooth. That is a real measurement I can make and you cannot, and it seems worth more than another round of arithmetic.

dknowles2:
Rebased cleanly — merging.

Checked the result rather than assuming: all four tests are present in `tests/test_init.py`, #351's own changes against `main` are exactly the three files it started with, and CI is green across build, mypy, ruff, Hassfest and HACS. The approval above still stands unchanged.

Thanks for taking the rebase back. Sorry again for the stray `feat/adaptive-polling` ref I left on the upstream repo — David has it flagged, and this PR's history was never touched by it.

PR #354, opened 2026-08-03
Resolve the grill spec on the advertised control board

### LINKED ISSUE #352: Setup resolves dual-board models on the wrong control board
The config flow derives the control board from the device-id prefix and filters the model list with it (`config_flow.py`: `grills.get_grills(control_board=...)`) — but setup then throws that information away. `async_setup_entry` constructs `PitBoss(conn, model, password=password)` without `control_board`, so pytboss resolves the spec by model name alone, which returns the variant the vendor lists most recently.

Seven models were sold on two board generations, and for every grill on the older board the resolved spec is the wrong one:

| Model | advertised board | board setup resolves |
|---|---|---|
| PB1150DX, PB850DX | PBL2 | PBL3 |
| PB850CS2, PB850M, PB1500NXW | PBM | PBM2 |
| PB1600PS2 | PBL | PBP |
| PBV3NX | PBV | PBV2 |

Concrete repro, straight from `grills.json`:

1. A grill advertising `PBL2-…` offers exactly `PB1150DX` / `PB850DX` in the picker — the flow filtered on PBL2, so the choice is legitimate.
2. Setup calls `get_grill("PB1150DX")` with no board → the PBL3 variant.
3. PBL2 and PBL3 differ precisely in whether the temperature routine converts to Celsius itself (`ftoc` is live in PBL3's routine and commented out in PBL2's — pytboss's `converts_temperatures_to_celsius` reads this distinction off the live code). A PBL2 grill working in Celsius is parsed with the PBL3 routine and reads wrong temperatures.

The board pairs on the other five models differ more broadly than one conversion block, so those grills get whatever the newer board's routines make of the older board's frames.

pytboss has carried the fix since 2026.8.2 — `PitBoss(..., control_board=...)` and `get_grill(name, control_board)` exist exactly for this, and their docstrings say to pass the advertised prefix. The integration just never passes it.

One migration wrinkle: under the board-remap workarounds this repo used to ship (#119, removed in #315), an entry could store a model that was never sold under its advertised prefix. Passing the prefix unconditionally would turn those entries into `InvalidGrill` at every setup, so the fix should fall back to today's by-name resolution (with a warning) when the pair does not exist.

Found while reviewing the integration against the pytboss catalogue; PR to follow.

trailro:
Fixes #352.

The config flow has always filtered the model picker by the device-id prefix, so the board is known and validated at entry creation — setup just never handed it to pytboss. This passes it through: `PitBoss(conn, model, password=password, control_board=device_id.split("-")[0])`.

For the seven dual-board models this is the difference between the right and the wrong parsing routines; for every single-board model it resolves to exactly the spec it did before, since the pair is unique.

**Fallback, deliberately:** an entry from the era of the board-remap workarounds (#119, removed in #315) can hold a model that was never sold under its advertised prefix. For those, `get_grill(model, prefix)` is checked first and on `InvalidGrill` the board is dropped with a warning, so the entry resolves by name alone — bit-for-bit what every release so far did. No existing install changes behaviour except the dual-board ones being fixed.

Per REVIEW.md this stays on the right side of the boundary: no board is remapped, nothing about frames or models lives here — the advertised prefix is passed through verbatim and pytboss does the resolving.

Test-wise: `conftest.py` splits the `mock_pitboss` fixture so the patched class is available for constructor assertions (`mock_pitboss_cls`), the instance fixture is unchanged. Two new tests: the prefix is passed through when the pair exists, and the fallback engages (with the warning) when it does not.

Verification: full suite green on Linux (123 passed — this environment is macOS, where the suite cannot run natively per AGENTS.md, so it ran in a `python:3.14` container instead); `ruff check`, `ruff format --check` and `mypy .` clean. I cannot verify on hardware — my grill is a PBL `PB1600PS1`, not a dual-board model — but the dual-board resolution difference is reproducible from `grills.json` alone and is quoted in the issue.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -23,7 +23,8 @@ from homeassistant.exceptions import ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 from homeassistant.helpers.device_registry import DeviceInfo
 from homeassistant.helpers.typing import ConfigType
-from pytboss import api, ble, wss
+from pytboss import api, ble, grills, wss
+from pytboss.exceptions import InvalidGrill
 from pytboss.transport import Transport
 
 from .const import (
@@ -116,8 +117,27 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     else:
         raise ValueError(f"Unknown protocol: {protocol}")
 
+    # The prefix the grill advertises names its control board, and a few
+    # models were sold on two board generations that do not parse alike.
+    # Resolving by model name alone picks the board the vendor lists most
+    # recently, which is the wrong one for every grill on the older board.
+    control_board: str | None = device_id.split("-")[0]
+    try:
+        await hass.async_add_executor_job(grills.get_grill, model, control_board)
+    except InvalidGrill:
+        # An entry from the era of board remapping can hold a model that was
+        # never sold under the advertised prefix. Resolving by name alone is
+        # what every release so far did, so those keep working unchanged.
+        LOGGER.warning(
+            "Model %s has no variant on control board %s; "
+            "using the vendor's latest listing for it",
+            model,
+            control_board,
+        )
+        control_board = None
+
     pitboss = await hass.async_add_executor_job(
-        lambda: api.PitBoss(conn, model, password=password)
+        lambda: api.PitBoss(conn, model, password=password, control_board=control_board)
     )
     device_info = DeviceInfo(
         identifiers={(DOMAIN, device_id)},
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -120,9 +120,10 @@ def mock_wss_conn() -> Generator[Mock]:
 
 
 @pytest.fixture
-def mock_pitboss(spec: Grill) -> Generator[Mock]:
-    with patch("pytboss.api.PitBoss", autospec=True) as mock_pitboss_cls:
-        api = mock_pitboss_cls.return_value
+def mock_pitboss_cls(spec: Grill) -> Generator[Mock]:
+    """The patched PitBoss class, for asserting on constructor arguments."""
+    with patch("pytboss.api.PitBoss", autospec=True) as mock_cls:
+        api = mock_cls.return_value
         api.spec = spec
         # `config` and `fs` are set in PitBoss.__init__, so autospec does not
         # know about them.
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
**This fixes more than the issue says it does, and the two extra consequences are worth putting on the record because neither is obvious from the diff.**

Seven dual-board models, confirmed by executing rather than reading:

```
PB1150DX  PBL2/PBL3     PB850DX   PBL2/PBL3
PB1500NXW PBM/PBM2      PB850CS2  PBM/PBM2
PB1600PS2 PBL/PBP       PB850M    PBM/PBM2
PBV3NX    PBV/PBV2
```

All four board pairs differ in `status_function`, and two also differ in their commands. Then I went looking for what that means for a user:

**1. A `PBV3NX` on the older `PBV` board silently loses entities today.** Resolving by name alone picks `PBV2`, which declares **none** of `turn-light-on`, `turn-light-off`, `turn-primer-motor-on` or `turn-primer-motor-off` — `PBV` declares all four. So those grills have had no light and no primer switch, and nothing anywhere would explain why.

**2. `PBL2` grills are currently being sent the wrong setpoint, and pytboss#541 made that worse rather than better.**

```
PB1150DX set-temperature(100, celsius):
  PBL2 (actual board)  -> FE0501010000FF   = 100
  PBL3 (what we pick)  -> FE0501020100FF   = 210
```

`PBL3` carries the unit-flag branch; `PBL2` does not. Before #541 both spelled it the same way, so the misresolution was harmless for temperatures. Since #541 a `PB1150DX` or `PB850DX` on a `PBL2` board running in Celsius has been asking for 100 and getting 210. I reviewed #541 and called it correct, which it is for the nine boards it targets — I did not think about it landing on top of a board we resolve wrongly.

That also reconciles something that looked contradictory: `PBL2` has zero models via `get_grills()`, because that yields one entry per model and always the later board. It is only reachable as an explicit `control_board` argument — which is exactly what this PR starts passing.

## The rest

* The fallback does what it claims: `get_grill('PB1150DX', 'PBV')` and `(..., 'NOTABOARD')` both raise `InvalidGrill`, and every single-board pair resolves to the same spec it did before, so no existing install changes except the seven.
* Per REVIEW.md this is on the right side of the line — the advertised prefix is passed through verbatim and pytboss does the resolving. No board is remapped here, which is the mistake this repo has made twice.
* Splitting `mock_pitboss` into class and instance fixtures is the right way to assert on constructor arguments without changing what every existing test sees.

**Running the suite in a `python:3.14` container is a genuinely useful find** and better than what I have been doing — I have been treating "cannot run locally on macOS" as "CI is the only check" and saying so on every review. I will use that.

One thing I did not verify: no dual-board grill exists to test against, yours included. The resolution difference is reproducible from `grills.json`, which is what makes the two consequences above checkable — but nobody has watched a `PBV3NX` on a `PBV` board gain its light switch back.

Labelled `bug` rather than `enhancement`: silently missing entities and a wrong setpoint are not a feature gap.

Merging.

PR #355, opened 2026-08-03
Create the firmware sensor up front and retry the failed read

### LINKED ISSUE #353: One failed firmware read at setup and the sensor does not exist until a reload
The firmware version is read once, in the coordinator's `_async_setup`, and deliberately never fails setup — but `sensor.py` only creates `FirmwareSensor` when that one read succeeded:

```python
if coordinator.firmware_version:
    entities.append(FirmwareSensor(coordinator, entry.unique_id))
```

So a single slow or dropped `PB.GetFirmwareVersion` answer — easy over a busy Bluetooth link, and exactly the moment setup happens to catch the board clearing its buffers — and the entity is simply absent. Nothing retries the read; the sensor stays missing until the integration is reloaded. The failure is invisible too: the fetch error is logged at debug level only, so what a user sees is a device that sometimes has a firmware sensor and sometimes does not, depending on how one RPC went at startup.

`test_no_firmware_sensor_when_the_grill_does_not_answer` currently pins this behaviour in, so it looks intended — it is not; it just codified what the first implementation did (I wrote it, mea culpa).

The shape that matches the rest of the integration: create the sensor unconditionally, let it report unavailable until a read succeeds (`available` already keys off `firmware_version is not None`), and have the coordinator retry the read on each poll until one lands — a no-op afterwards, since the version cannot change outside a reload. That mirrors how `sys_info` is refreshed: never fatal, and self-healing.

PR to follow.

trailro:
Fixes #353.

`FirmwareSensor` is now created unconditionally. It reports unavailable until a read succeeds — `available` already keyed off `firmware_version is not None`, so the entity needed no change beyond its comment — and the coordinator retries the read on each poll until one lands, after which it is a no-op: the version cannot change outside a reload.

The retry lives in `_async_refresh_firmware_version`, shaped exactly like `_async_refresh_sys_info` next to it: never fatal, debug-logged on failure. `_async_setup` calls the same method, so the setup path is unchanged apart from the extraction.

`test_no_firmware_sensor_when_the_grill_does_not_answer` pinned the old behaviour in, so it is replaced rather than kept: the new test fails the first read, asserts the sensor exists and is unavailable, then lets the next poll's read succeed and asserts the version appears. The happy-path test (`test_firmware_version_sensor`, including the device `sw_version` assertion) is untouched and still passes.

Known limitation, deliberately out of scope: when the version arrives only after setup, the device-registry `sw_version` field still waits for the next reload — that write happens in `async_setup_entry` and moving it into the coordinator drags device-registry access in there for a purely cosmetic field. The sensor itself is correct either way.

Verification: full suite green on Linux (121 passed — this environment is macOS, where the suite cannot run natively per AGENTS.md, so it ran in a `python:3.14` container instead); `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -166,11 +166,23 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         """Set up the coordinator."""
         await self.api.subscribe_state(self._on_state_update)
         await self._start_api()
+        await self._async_refresh_firmware_version()
+
+    async def _async_refresh_firmware_version(self) -> None:
+        """Fetch the firmware version until one read succeeds. Never fatal.
+
+        The version does not change outside a reload, so this is a no-op once
+        known. Retrying until then matters because the read at setup used to
+        be the only attempt: a grill that failed to answer it once had no
+        firmware sensor until the integration was reloaded.
+        """
+        if self.firmware_version is not None:
+            return
         try:
             result = await self.api.get_firmware_version()
             self.firmware_version = result.get("firmwareVersion")
         except Exception as ex:  # noqa: BLE001
-            # Cosmetic; never worth failing setup over.
+            # Cosmetic; never worth failing a refresh over.
             self.logger.debug("Could not fetch the firmware version: %s", ex)
 
     def _merge_state(self, state: StateDict) -> StateDict:
@@ -250,6 +262,7 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
             raise UpdateFailed("Grill not connected") from ex
 
         await self._async_refresh_sys_info()
+        await self._async_refresh_firmware_version()
 
         # Always fetch the current state to ensure sensors stay up-to-date.
         # Relying solely on push notifications means sensors can go stale after
--- a/custom_components/pitboss/sensor.py
+++ b/custom_components/pitboss/sensor.py
@@ -93,8 +93,7 @@ async def async_setup_entry(
         entities.append(ProbeSensor(coordinator, entry.unique_id, entity_description))
     for description in SYS_INFO_DESCRIPTIONS:
         entities.append(SysInfoSensor(coordinator, entry.unique_id, description))
-    if coordinator.firmware_version:
-        entities.append(FirmwareSensor(coordinator, entry.unique_id))
+    entities.append(FirmwareSensor(coordinator, entry.unique_id))
     if coordinator.api.spec.json.get("has_recipe_functionality", False):
         for entity_description in RECIPE_ENTITY_DESCRIPTIONS:
             entities.append(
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
Right fix, and the entity needed no change because `available` was already keyed on the version being known — that is the sign the original bug was in the wrong place. Creating the sensor conditionally meant one failed read at setup removed the entity permanently, and no amount of retrying would have brought it back.

Replacing `test_no_firmware_sensor_when_the_grill_does_not_answer` rather than keeping it is right, and calling that out is what a reviewer needs: it encoded the behaviour being fixed, so leaving it would have meant the fix and the test disagreed about what correct means.

The new test covering the whole sequence — fail, entity exists and is unavailable, next poll succeeds, version appears — is the shape that matters, because the interesting part is the transition rather than either end of it.

## One asymmetry with the method it is modelled on

> The retry lives in `_async_refresh_firmware_version`, shaped exactly like `_async_refresh_sys_info` next to it.

Almost — `_async_refresh_sys_info` is **time-gated**:

```python
now = monotonic()
if self.sys_info and now - self._sys_info_at < SYS_INFO_INTERVAL:
    return
```

This one gates only on success. Once the version is known they behave the same, and on the happy path that is after one read. But a grill that persistently fails `PB.GetFirmwareVersion` gets a retry on **every** poll forever — and since #351 that is every 10 seconds while the grill is lit, on Bluetooth, indefinitely.

Probably harmless: the read is cheap and a grill that never answers it is unusual. But it is the one case where the two methods diverge, and "shaped exactly like" is what would stop the next person noticing. A `_firmware_attempted_at` gate, or simply giving up after a few tries, would close it.

Not blocking — the change is correct as written and the failure mode is extra traffic on an already-broken grill, not wrong data. Flagging so it is a decision rather than an oversight.

Agreed on leaving the device-registry `sw_version` alone. Dragging device-registry access into the coordinator for a cosmetic field that corrects itself on the next reload is a poor trade, and saying so explicitly is better than silently leaving it.

CI green: build, mypy, ruff, Hassfest, HACS. Labelled `bug`. Merging.

trailro:
Thanks for the review — and for catching the asymmetry. You are right that "shaped exactly like" oversold it: the sys-info read is time-gated and this one was gated on success alone, so a grill that persistently fails `PB.GetFirmwareVersion` would have been asked on every 10-second poll while lit. That is exactly the kind of divergence worth being a decision rather than an oversight, so: #356 gates retries to `SYS_INFO_INTERVAL`, same cadence and same rationale as the read it was modelled on. One wrinkle documented there: the attempt timestamp starts as `None` rather than `0.0`, because `monotonic()` can be small at process start and `0.0` would gate the *first* attempt — the sys-info read dodges that by content-checking first, which a not-yet-known version cannot.

It is stacked on this branch, so the diff clears to just the gate once this lands.

PR #356, opened 2026-08-03
Gate firmware retries to the diagnostics cadence

trailro:
Follow-up to the review of #355, taking the flagged asymmetry as a decision: retries now follow the diagnostics cadence rather than the poll rate.

An attempt timestamp gates `_async_refresh_firmware_version` to `SYS_INFO_INTERVAL`, the same cadence as the system-info read it was modelled on — reusing that constant rather than minting a new one, since the rationale is identical: diagnostics are not worth a round trip on every poll. A grill that persistently fails the read is now asked at most once a minute instead of on every 10-second poll while lit. The happy path is unchanged: one read, then the success gate makes everything else a no-op.

One deliberate difference from `_async_refresh_sys_info`: the timestamp starts as `None` rather than `0.0`. `monotonic()` can be small at process start, and `0.0` would read as "just tried" and gate the *first* attempt — the sys-info read dodges this by content-checking `self.sys_info` first, which a not-yet-known version cannot do.

The recovery test now covers the gate from both sides: a poll landing inside the interval does not ask again (asserted on the mock's await count, with the entity still unavailable), and once the gate elapses the next poll's read succeeds and the version appears. Monotonic time cannot be frozen from a test, so the gate is aged by adjusting the timestamp directly.

Was stacked on #355; now that it has landed, `main` is merged in (a merge rather than a rewrite, to keep the pushed history append-only) and the diff is just the gate: `coordinator.py` +16, `test_sensor.py` +26.

Verification: full suite green on Linux (123 passed, in a `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check` and `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -56,6 +56,10 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         # Latest Sys.GetInfo payload from the control board.
         self.sys_info: dict = {}
         self._sys_info_at = 0.0
+        # When the firmware version was last asked for. `None` rather than
+        # 0.0 so the first attempt is never gated: `monotonic()` can be
+        # small at process start, which 0.0 would read as "just tried".
+        self._firmware_attempted_at: float | None = None
         # What the grill is holding, and what we are holding for it. See
         # `probe_target` for why both exist.
         self.probe_targets: dict[int, int] = {}
@@ -175,9 +179,21 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         known. Retrying until then matters because the read at setup used to
         be the only attempt: a grill that failed to answer it once had no
         firmware sensor until the integration was reloaded.
+
+        Retries are gated to the diagnostics cadence, like the system info
+        read above. Gating on success alone meant a grill that persistently
+        failed this read was asked again on every poll -- every ten seconds
+        for as long as it was lit, once the poll interval followed the grill.
         """
         if self.firmware_version is not None:
             return
+        now = monotonic()
+        if (
+            self._firmware_attempted_at is not None
+            and now - self._firmware_attempted_at < SYS_INFO_INTERVAL
+        ):
+            return
+        self._firmware_attempted_at = now
         try:
             result = await self.api.get_firmware_version()
             self.firmware_version = result.get("firmwareVersion")
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -166,11 +166,23 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         """Set up the coordinator."""
         await self.api.subscribe_state(self._on_state_update)
         await self._start_api()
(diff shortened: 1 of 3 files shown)


dknowles2:
Substance is right, and **the `None`-versus-`0.0` detail is a better catch than the thing I flagged.**

I checked both halves of it. `_async_refresh_sys_info` gates on `if self.sys_info and now - self._sys_info_at < SYS_INFO_INTERVAL` — the content check comes first, so on the very first call the empty `sys_info` short-circuits it and the `0.0` timestamp is never consulted. A firmware version cannot do that, because `None` *is* its not-yet-known state and it is already the success gate.

And the failure it avoids is real rather than theoretical:

```
monotonic() on this host = 1060146s   (12 days up, so 0.0 would be harmless here)
```

But on a host booted seconds ago — a Pi coming back after a power cut, which is exactly when a grill read is most likely to fail — `monotonic()` is small, `now - 0.0 < 60` holds, and the **first** attempt would be gated. A bug that only appears on a freshly booted system and clears itself within a minute is close to undiagnosable. Choosing `None` and saying why in the description is worth more than the gate itself.

Reusing `SYS_INFO_INTERVAL` rather than minting a constant is right: the rationale really is identical, and a second constant would invite the two to drift.

## It needs the rebase you planned

#355 landed a few minutes ago, so this is now **DIRTY** — `coordinator.py` and `tests/test_sensor.py` both conflict, since the stacked commit and `main` now carry the same content. CI has not run either; only `automerge: skipped` is present.

Rebasing onto `aa964d3` should reduce this to the gate alone, as you said it would. I am deliberately not touching your branch — I pushed to the wrong remote trying to be helpful with #351 earlier and left a stray ref on the upstream repo for it, so that lesson is learned.

Happy to merge once it is rebased and green; the review above stands and I do not expect to have anything further.

Labelled `bug` — sustained pointless traffic on a BLE link that has been observed dropping for half an hour at a time is nearer a defect than hygiene, even though #355 was not wrong to land without it.

One aside, since it changes how I work rather than how you do: running the suite in a `python:3.14` container is a better answer than what I have been doing. I have been reporting "cannot run locally on macOS, CI is the real check" on every ha-pitboss review, which is true and also a smaller effort than was available.

dknowles2 (approved):
Rebased cleanly — the diff against `main` is now the gate alone: sixteen lines in `coordinator.py` and the two-sided test. CI green across build, mypy, ruff, Hassfest and HACS.

The review I left before the rebase stands, so only the short version here: **the `None`-versus-`0.0` choice is the part worth keeping**, and the comment on the field says why in two lines rather than leaving the next reader to wonder why it is not `0.0` like its neighbour. I checked both halves — `_async_refresh_sys_info` gets away with `0.0` only because its content check on `self.sys_info` short-circuits first, and a firmware version cannot do that because `None` is already its success gate.

That the resulting bug would have been invisible on any host with meaningful uptime, and would have cleared itself within a minute on the ones where it appeared, is what makes it worth having caught in review rather than in the field.

Gating both sides of the interval in the test — a poll inside it does not re-ask, one after it does — is the right shape. Aging the timestamp directly rather than trying to freeze `monotonic()` is the pragmatic call and you flagged it as such.

Merging.

PR #358, opened 2026-08-03
Make the connectivity sensor follow the grill, not just the socket

### LINKED ISSUE #357: The connectivity sensor reports the relay socket, not the grill, on the cloud protocol
`ConnectivitySensor.is_on` is `api.is_connected()`. On Bluetooth that is faithful — the link is to the grill itself. On the cloud protocol it is not: `WebSocketConnection.is_connected()` reflects the socket to `socket.dansonscorp.com`, and the relay accepts that connection whether or not a grill is behind it (it accepts `/to/<id>` for arbitrary ids; there is no handshake with the grill). So with `wss` selected, a powered-off or unreachable grill reads **Connected** — while the class docstring promises "whether the grill is currently reachable", and the sensor's whole purpose (from #330) is to be the stable signal for grill-offline automations.

The transport cannot know more than its own socket, so this is not pytboss's to fix — whether the grill answers polls is the liveness signal, and only the coordinator has that. Fix: require both, `api.is_connected() and coordinator.last_update_success`.

Consequences, stated plainly:
- On Bluetooth, effectively unchanged: a connected grill answers its pings, and a dropped link already flips `is_connected()`.
- On the cloud protocol the sensor becomes truthful, at poll granularity: a grill that dies stays "Connected" for up to one poll interval plus the 10s ping timeout before flipping. Reachability through an always-up relay was never going to be instant; wrong-but-instant is the thing being removed.
- Same shape holds for any future transport (#344 / pytboss#543 are adding a third), which is the other reason to compose it at the consumer: `is_connected()` means something slightly different per transport, `last_update_success` means the same thing on all of them.

Caveat on evidence: the relay's accept-anything behaviour is established from the firmware source and by connecting to the relay directly, not from a live powered-off-grill session over `wss`. The code-level half — `is_connected()` reads only `self._sock` — needs no experiment.

PR to follow.

trailro:
Fixes #357.

`is_on` now requires `coordinator.last_update_success` alongside `api.is_connected()`: the transport can only vouch for its own socket, and on the cloud protocol that socket is to the vendor's relay, which accepts the connection whether or not the grill is behind it. Whether the grill answers polls is the liveness signal, and only the coordinator has it — which is why this composes here rather than in pytboss.

On Bluetooth this changes nothing in practice (a connected grill answers its pings; a dropped link already flips `is_connected()`). On the cloud protocol the sensor stops reporting a powered-off grill as Connected, at the cost of detection moving to poll granularity — up to one poll interval plus the 10s ping timeout. It also gives the sensor one meaning across transports, which matters as #344 / pytboss#543 add a third: `is_connected()` differs per transport, `last_update_success` does not.

New test drives exactly the lying case: transport still connected, polls start failing (`async_set_update_error`), sensor goes off. The existing socket-drop test is untouched and still passes.

Class docstring records the reasoning, since "why is this not just `is_connected()`" is the first question the code now raises.

Verification: full suite green on Linux (124 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check`, `mypy .` clean. The relay's accept-anything behaviour is established from the firmware source and direct connections to the relay, not a live powered-off-grill `wss` session — the caveat is in #357.

### DIFF
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -164,6 +164,12 @@ class ConnectivitySensor(BaseEntity, BinarySensorEntity):
     Unlike every other entity this one stays available while the grill is
     disconnected: an entity that disappears cannot say the grill is offline,
     which is the one thing an offline automation needs to hear.
+
+    Reachable means the transport is connected *and* the grill answers
+    polls. The transport alone is not enough: on the cloud protocol
+    `is_connected()` reflects the socket to the vendor's relay, which
+    accepts the connection whether or not the grill is on the other side
+    of it -- so a powered-off grill read as "connected".
     """
 
     _attr_device_class = BinarySensorDeviceClass.CONNECTIVITY
@@ -184,7 +190,11 @@ class ConnectivitySensor(BaseEntity, BinarySensorEntity):
 
     @property
     def is_on(self) -> bool:
-        return bool(self.coordinator.api) and self.coordinator.api.is_connected()
+        return (
+            bool(self.coordinator.api)
+            and self.coordinator.api.is_connected()
+            and self.coordinator.last_update_success
+        )
 
 
 class MeatProbeControlSensor(BaseEntity, BinarySensorEntity):
--- a/tests/test_binary_sensor.py
+++ b/tests/test_binary_sensor.py
@@ -114,6 +114,34 @@ async def test_connectivity_reports_offline_instead_of_vanishing(
     assert state.state == "off"
 
 
+async def test_connectivity_follows_the_grill_not_just_the_socket(
+    hass: HomeAssistant,
+    mock_add_config_entry: Callable[[], Awaitable[MockConfigEntry]],
+    mock_pitboss: Mock,
+) -> None:
+    """A connected transport whose grill stops answering is not reachable.
+
+    On the cloud protocol `is_connected()` reflects the socket to the
+    vendor's relay, which stays open whether or not the grill is behind
+    it -- a powered-off grill read as "connected" until this check also
+    asked whether polls succeed.
+    """
+    entry = await mock_add_config_entry()
+    coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
Right split, and the argument for why it lives here rather than in pytboss is the part that matters: `is_connected()` can only ever vouch for the socket, and on the cloud protocol that socket terminates at the vendor's relay. Only the coordinator knows whether anything answered. Putting it in the library would mean the library needing a poller, which is the same boundary question I worked through on pytboss#543 and came down on the same side of.

**And it gets more valuable as transports multiply, not less.** `is_connected()` already means three different things — a BLE link, a relay socket, and (with pytboss#543) whether an HTTP host last answered. `last_update_success` means the same thing under all three. A sensor that reads "Connected" differently depending on how you set the grill up is worse than one that is a poll interval late.

Being explicit that detection moves to poll granularity — up to one interval plus the 10s ping timeout — is the honest trade, and #351 made that interval 60s in standby, which is when a grill is most likely to be off. So worst case is around 70 seconds to notice a powered-off grill on the cloud protocol. Still better than never.

## Checked

* `DataUpdateCoordinator.last_update_success` initialises to `True`, so there is no false "disconnected" flash between setup and the first refresh.
* The condition is strictly narrowing: `_async_update_data` already raises `UpdateFailed` when `is_connected()` is false, so this can only ever turn an on into an off, never the reverse.
* CI green.

## One semantic edge, not a request

Since #350, a rejected password raises `ConfigEntryAuthFailed`, which also clears `last_update_success` — so a grill that is powered on, reachable and answering will read **Disconnected** when the stored password has gone stale. Arguably wrong in the narrow sense, arguably right in the useful one, since Home Assistant genuinely cannot talk to it and there is a reauth flow in front of the user either way.

Not worth a special case; the reauth prompt is the louder signal. Recording it so that if someone reports "Connectivity says offline but the grill is on", the password is the first thing to check.

Saying plainly that the relay's accept-anything behaviour comes from the firmware source and direct relay connections rather than a live powered-off `wss` session is the right framing, and it is the same standard I have been asking of myself. Labelled `bug`. Merging.

PR #359, opened 2026-08-03
Sensor and setup housekeeping

trailro:
Three small cleanups from a pass over the sensor platform and setup path. None changes what a working grill reports; no issue filed since none has a user-visible failure mode worth one.

**`recipeTime` loses its `MEASUREMENT` state class.** It is a countdown; long-term statistics of "seconds remaining" mean nothing and pollute the statistics tables. The dataclass default stays for `recipeStep`; the description overrides it with `None`. One consequence to know about: entities that already recorded statistics have them orphaned, and Home Assistant shows the standard repair offering to drop them — that is the intended outcome, not a regression.

**`RecipeSensor.available` read a missing `moduleIsOn` as on.** The power switch reads the same key as off (`data.get("moduleIsOn", False)`). Off wins for consistency — a frame that does not carry the key should not read as "the grill is on" in one place and "off" in another. Reachable only on a partial frame before the merged state has seen the key once, so in practice this is about the two code paths agreeing rather than observable behaviour.

**The `PitBoss` constructor is built inline.** The executor job around it was a holdover from when the constructor read the grill catalogue; spec resolution moved into `start()` long ago and the constructor only assigns attributes now. The `get_grill` validation call above it stays in the executor — that one does touch the catalogue.

Both behaviour tweaks are pinned by tests: `recipe_time` state carries no `state_class` attribute, and a frame without `moduleIsOn` reads unavailable.

Verification: full suite green on Linux (124 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -136,9 +136,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         )
         control_board = None
 
-    pitboss = await hass.async_add_executor_job(
-        lambda: api.PitBoss(conn, model, password=password, control_board=control_board)
-    )
+    # Constructed inline: since spec resolution moved into `start()`, the
+    # constructor only assigns attributes. The executor job it ran in was a
+    # holdover from when it read the grill catalogue.
+    pitboss = api.PitBoss(conn, model, password=password, control_board=control_board)
     device_info = DeviceInfo(
         identifiers={(DOMAIN, device_id)},
         name=device_id,
--- a/custom_components/pitboss/sensor.py
+++ b/custom_components/pitboss/sensor.py
@@ -58,7 +58,7 @@ class RecipeSensorEntityDescription(SensorEntityDescription):
     """Describes a PitBoss recipe sensor."""
 
     key: Literal["recipeTime", "recipeStep"]
-    state_class: SensorStateClass = SensorStateClass.MEASUREMENT
+    state_class: SensorStateClass | None = SensorStateClass.MEASUREMENT
 
 
 RECIPE_ENTITY_DESCRIPTIONS = (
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
Three unrelated small things, each with its own justification — and I verified all three rather than taking the bundle as read.

**The constructor really is inert now.** `get_grill` appears exactly once in `pytboss/api.py`, at line 126, inside `start()` behind `asyncio.to_thread`. `__init__` assigns attributes and builds `FileSystem` / `Config` / `OTA` wrappers. So the executor job was pure ceremony, and keeping the one around the `get_grill` validation above it is the right line to draw — that one does read the catalogue and does run dukpy.

**`moduleIsOn` defaulting to `True` was the odd one out**, confirmed across the tree: `climate.py` reads it as `data.get("moduleIsOn", False)` in three places, and `coordinator.py` uses a bare `.get()` which is falsy-by-default in both the seeding gate and the poll-interval decision. `RecipeSensor` was alone in assuming on. Agreeing on the default matters more than which default, and off is the one everything else already uses.

**Dropping `MEASUREMENT` from `recipeTime` is right and the consequence is worth having flagged.** Long-term statistics of a countdown are noise, and `DURATION` with `SECONDS` still gives the nice display. The orphaned-statistics repair prompt is the correct outcome rather than a regression — but a user seeing it will not know that, so it being in the PR description means the answer exists somewhere when they ask.

Widening the dataclass annotation to `SensorStateClass | None` rather than dropping the default is the light-touch version: `recipeStep` keeps `MEASUREMENT` without restating it.

"No issue filed since none has a user-visible failure mode worth one" is the right call too. Two of these are invisible on a working grill and the third is a statistics-hygiene decision.

CI green. Labelled `cleanup`. Merging.

PR #360, opened 2026-08-03
Build the config flow's model list off the event loop

trailro:
`_show_more_info_form` called `grills.get_grills()` inline:

```python
models = [g.name for g in grills.get_grills(control_board=control_board)]
```

The first such call reads and parses `grills.json` — about 180 kB — and then builds a `Grill` for every model on that board, each of which runs the JS-scrubbing regexes. `async_setup_entry` already hands the equivalent `get_grill()` call to the executor; this path was missed.

It is also the path that pays the cost. `_get_grills()` is `@cache`d, but on a fresh install there is no entry yet whose setup could have warmed it, so the config flow is exactly where the parse happens and where Home Assistant's blocking-call detector fires.

## The change

The lookup moves to a module-level helper handed to `async_add_executor_job`, which makes `_show_more_info_form` a coroutine. Both callers were already `async`, so they just gained an `await`.

## Testing

The new test records which thread `get_grills` runs on and asserts it is not the event loop's. It **fails on the unpatched code** — I checked by restoring `config_flow.py` from `main` and keeping the test.

126 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

One note on how I ran them: the suite does not start on macOS — setting up the `bluetooth` component fails in `dbus_fast` with `'NoneType' object is not callable`, which takes 110 of the 125 tests with it. I ran everything in a Linux container instead, matching what CI does, and confirmed the baseline was green there before changing anything.

### DIFF
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -15,6 +15,14 @@ from pytboss import grills
 from .const import ALL_PROTOCOLS, DEFAULT_PROTOCOL, DOMAIN, LOGGER
 
 
+def _models_on_board(control_board: str) -> list[str]:
+    """Model names this control board is sold under.
+
+    A module-level function so it can be handed to the executor.
+    """
+    return [g.name for g in grills.get_grills(control_board=control_board)]
+
+
 class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
     """Config flow for PitBoss."""
 
@@ -76,9 +84,9 @@ class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
                     CONF_PROTOCOL: user_input.get(CONF_PROTOCOL, DEFAULT_PROTOCOL),
                 },
             )
-        return self._show_more_info_form("more_info")
+        return await self._show_more_info_form("more_info")
 
-    def _show_more_info_form(
+    async def _show_more_info_form(
         self,
         step_id: str,
         model: str | vol.Undefined = vol.UNDEFINED,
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Right fix, and the precedent you cite is real — `__init__.py:126` already hands `get_grill` to the executor for exactly this reason, so this path was simply missed rather than deliberately different.

I measured the cost rather than taking the file size as the argument:

```
grills.json                179,899 bytes
first get_grills() call      7 ms (cold cache), 7 models on PBL
second call                4.1 ms (warm)
```

**7 ms is modest, and the duration is not really the point** — `_get_grills()` loads through `resources.files(...).read_text()`, and Home Assistant patches file I/O in the event loop and reports it whatever it costs. So the detector fires on the read itself. On a Pi reading from SD rather than this laptop's page cache, the number would also be a good deal worse.

**Your point about *where* the cost lands is the one I would not have thought of.** `_get_grills()` is `@cache`d, so it looks like a one-off — but on a fresh install there is no entry yet whose setup could have warmed it, which means the config flow is not just *a* caller, it is reliably *the* caller that pays. A cache makes this look harmless everywhere except the one place it isn't.

A module-level `_models_on_board` rather than a lambda or a partial is the right shape for `async_add_executor_job`, and making `_show_more_info_form` a coroutine cost nothing since both callers were already async.

## Verified

* **I could not run the tests.** `tests/test_config_flow.py` fails 15/15 on unmodified `main` here, so it is environmental as always and CI is the real check. I did not reproduce your thread-affinity test.

**Your diagnosis of the macOS failure is better than mine** and I am going to use it: `dbus_fast` raising `'NoneType' object is not callable` while setting up the `bluetooth` component, taking 110 of 125 tests with it. I have been describing the symptom I happened to see downstream of that. Running in a Linux container and confirming the baseline green *before* changing anything is also the right discipline — it distinguishes "my change broke this" from "this was already broken", which is the whole difficulty with a suite that cannot run natively.

Labelled `bug` — blocking the event loop is a defect rather than hygiene, even though nothing visibly misbehaves. Merging.

PR #361, opened 2026-08-03
Hold a probe target we set until the grill reports it

trailro:
Setting a probe target makes the slider jump back to its previous value for up to a poll interval.

## Why

`coordinator.probe_target()` reads the board's own `pNTarget` first, which is right — a target set from the phone shows up immediately that way. But it is still the *previous* value for as long as it takes the board to report the new one, and `async_set_native_value` ends with `async_update_listeners()`, so the entity re-read the old target and rendered it.

Not an edge case. I checked the catalogue with comments stripped — a plain grep over `grills.json` says all 20 boards, but five have the line commented out — and **15 of the 20 boards report `p1Target` in live code**: LBL, LFS, PBA, PBB, PBD, PBE, PBL, PBL2, PBL3, PBP, PBT, PBV, PBV2, PBVA, PBX1. So this is what most grills do, on the control probe.

## The change

The same shape `select.py` already uses, which the grill's temperature unit needed for the same reason: show what we asked for, clear it as soon as the coordinator reports that value, and give up on it after one settle window if the grill never does. `MCU_SETTLE_SECONDS` moves to `const.py` now that it has two callers.

`ProbeTargetReachedSensor` deliberately still follows the coordinator rather than the pending value — it answers whether the grill has reached its target, which is a question about the grill, not about what we asked for.

## Testing

Two tests, both **verified to fail on the unpatched code** (restored `number.py` from `main`, kept the tests): the value survives a refresh that still carries the old `p1Target`, and the pending value is dropped once the board confirms.

127 pass, `ruff check`, `ruff format --check` and `mypy .` clean.

As on #360: the suite does not run on macOS — the `bluetooth` component fails to set up in `dbus_fast` — so I ran everything in a Linux container matching CI, and confirmed the baseline was green there first.

### DIFF
--- a/custom_components/pitboss/const.py
+++ b/custom_components/pitboss/const.py
@@ -51,3 +51,11 @@ def probe_label(has_mpc: bool, probe_number: int) -> str:
     if not has_mpc:
         return f"Probe {probe_number}"
     return "MPC" if probe_number == 1 else f"P{probe_number}"
+
+
+MCU_SETTLE_SECONDS = 6
+"""Seconds to show a value we asked for before believing the grill again.
+
+The board wipes its cached status the moment it forwards a command to the
+MCU, so the next poll still carries the old value and an entity that read it
+straight back would appear to snap to the previous setting."""
--- a/custom_components/pitboss/number.py
+++ b/custom_components/pitboss/number.py
@@ -7,8 +7,9 @@ from homeassistant.components.number import NumberEntityDescription, RestoreNumb
 from homeassistant.components.number.const import NumberDeviceClass
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import UnitOfTemperature
-from homeassistant.core import HomeAssistant
+from homeassistant.core import CALLBACK_TYPE, HomeAssistant, callback
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
+from homeassistant.helpers.event import async_call_later
 from homeassistant.util.unit_conversion import TemperatureConverter
 
 from .const import (
@@ -17,6 +18,7 @@ from .const import (
     DEFAULT_PROBE_MAX_TEMP,
     DEFAULT_PROBE_MIN_TEMP,
     DOMAIN,
+    MCU_SETTLE_SECONDS,
     probe_label,
 )
 from .coordinator import PitBossDataUpdateCoordinator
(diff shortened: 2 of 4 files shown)


dknowles2 (approved):
The fix is right and it is the #347 pattern applied properly — including the lesson from that review, registering the removal callback once in `async_added_to_hass` rather than per set, and giving the pending value an escape after one settle window. Nothing to add on the code.

**But the board count is wrong, and it is wrong the same way #541's and #550's were.**

You were right that a raw grep over-counts and that comments are the reason. It just does not stop at five:

```
live p1Target in status_function :  3 boards  -> LBL, LFS, PBX1
commented out                    : 17 boards  -> PBA PBB PBC PBC2 PBD PBE PBG PBL
                                                 PBL2 PBL3 PBM PBM2 PBP PBT PBV PBV2 PBVA
mentioned nowhere                :  0
live p1Target in temperatures_function : 0 boards
```

So it is 3, not 15 — and your list includes `PBL`, `PBV`, `PBT` and nine others whose line is commented out. `PBX1` has no models, so **9 of 137 models are on a board that reports `p1Target` at all**: five on LBL, four on LFS.

I went looking because `_scrub_js` gave me 20 at first and I did not believe it. That turned out to be my own error — `_scrub_js` is applied to *command* functions, not to `status_function`, which dukpy executes with its comments intact. Stripping them myself is what gives the number above, and a behavioural check agrees: driving `parse_status` over every board, only LBL emits `p1Target` for a frame I could construct.

**None of that changes whether this should land.** For the 128 models whose board does not report it, `async_set_probe_target` puts the value into `coordinator.probe_targets` immediately, so `probe_target()` already returns the new one and the pending value clears on the first update — harmless. For the nine on LBL and LFS the snap-back is real and this fixes it.

What changes is the claim "this is what most grills do, on the control probe". It is what a small minority do. Worth correcting because the next person sizing a probe-target change will read that sentence and believe it — and because the same mistake has now produced a wrong number three times, each from reading `grills.json` rather than executing it.

`MCU_SETTLE_SECONDS` moving to `const.py` on its second caller is right, and the docstring there explains the mechanism rather than restating the constant.

And leaving `ProbeTargetReachedSensor` on the coordinator is the correct call for the reason you give: it answers a question about the grill, not about what we asked for.

CI green: build, mypy, ruff, Hassfest, HACS. **I could not run the tests** — the suite does not start here, as you note. Labelled `bug`. Merging.

PR #362, opened 2026-08-03
Put a late firmware read on the device page too

trailro:
`_async_refresh_firmware_version` exists because the read at setup used to be the only attempt — its docstring says it: *"a grill that failed to answer it once had no firmware sensor until the integration was reloaded"*.

But the device registry is still written exactly once, in `async_setup_entry`:

```python
if coordinator.firmware_version:
    coordinator.device_info["sw_version"] = coordinator.firmware_version
```

and `device_info` is only read when the device is created. So half the fix landed: when a retry finally succeeds, the sensor populates and the **device page stays blank forever** — for exactly the grill the retry was added for.

## The change

The coordinator publishes it now, in the one place that knows when it arrived. `device_info` still covers the first read, which happens before the device exists (the registry entry is created from it during platform setup); a read that only succeeds later updates the registry directly. The one-shot write in `async_setup_entry` goes, since it was the same thing done in one of the two cases.

## Testing

A new test drives the same path the existing recovery test does — first read fails, a later one succeeds — and asserts the registry, not the sensor. **It fails on the unpatched code**, where `device.sw_version` stays `None`.

127 pass, `ruff check`, `ruff format --check` and `mypy .` clean, run in a Linux container as on #360 (the suite does not start on macOS).

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -160,9 +160,6 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         await pitboss.stop()
         raise
 
-    if coordinator.firmware_version:
-        coordinator.device_info["sw_version"] = coordinator.firmware_version
-
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
     return True
 
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -7,6 +7,7 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.const import UnitOfTemperature
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryAuthFailed
+from homeassistant.helpers import device_registry as dr
 from homeassistant.helpers.device_registry import DeviceInfo
 from homeassistant.helpers.update_coordinator import DataUpdateCoordinator, UpdateFailed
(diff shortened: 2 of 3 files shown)


dknowles2 (approved):
**"Half the fix landed" is exactly right, and it is the sharpest kind of bug to find** — the retry machinery from #355 exists precisely for a grill that fails the first read, and for that grill the device page stayed blank forever anyway. The sensor populating was enough to make it look fixed.

The mechanism holds: the one-shot write sat *before* `async_forward_entry_setups`, so the first read makes it into the registry entry at creation. Mutating the same dict afterwards updates nothing, because nothing reads it again.

**Splitting it the way you have is the right shape rather than replacing one with the other.** `device_info` still covers the first read — which has to happen that way, since the device does not exist yet — and the registry update covers a read that lands later. The `async_get_device` returning `None` at setup time is not a guard against an error, it is the normal path for the first case, and the docstring says so.

Moving the write into the coordinator is also where it belongs: it is the only place that knows *when* the version arrived, which is the whole question.

**And asserting the registry rather than the sensor is what makes the test worth having.** The existing recovery test already covered the sensor, and passed while this was broken — testing the visible half of a two-part write is how the gap survived in the first place.

CI green: build, mypy, ruff, Hassfest, HACS. **I could not run the tests** — the suite does not start here — so CI is the check, and I did not reproduce your `device.sw_version is None` result on unpatched code.

Labelled `bug`. Merging.

PR #363, opened 2026-08-03
Stop retrying two failures that retrying cannot fix

trailro:
Two failures present as transient and are not. Both need the user, and neither gets to them today.

## An unknown model retries forever

Setup resolves the grill with the advertised prefix as its control board, and falls back to resolving by name alone when that pair does not exist — for entries the old board-remap workarounds left behind. But the fallback is never validated. If the model itself is not in the catalogue (a renamed entry after a definitions refresh, a hand-edited entry), `api.start()` raises `InvalidGrill` from inside the coordinator, where Home Assistant logs *"Unexpected error fetching pitboss data"* with a traceback and turns it into `ConfigEntryNotReady` — an escalating backoff, forever, for something no amount of waiting resolves.

The fallback is checked now, and a model that resolves neither way raises `ConfigEntryError`.

## A rejected password gives a toast and no way out

`set_grill_password` authenticates with the *current* password, so `Unauthorized` there means the one Home Assistant holds is not the grill's. That is precisely the condition the coordinator turns into `ConfigEntryAuthFailed` and a reauth flow — but the service wraps every exception into `HomeAssistantError`, so the user sees a flat error and is left to guess.

It now starts the reauth flow and says what happened. The entry is still left alone, as before.

## Testing

Two tests, both **verified to fail on the unpatched code**: an entry whose model is not in the catalogue ends in `SETUP_ERROR` rather than `SETUP_RETRY`, and a rejected password leaves a reauth flow in progress.

128 pass, `ruff check`, `ruff format --check` and `mypy .` clean, run in a Linux container as on #360 and #362.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -19,7 +19,7 @@ from homeassistant.const import (
     Platform,
 )
 from homeassistant.core import HomeAssistant, callback
-from homeassistant.exceptions import ConfigEntryNotReady
+from homeassistant.exceptions import ConfigEntryError, ConfigEntryNotReady
 from homeassistant.helpers.aiohttp_client import async_get_clientsession
 from homeassistant.helpers.device_registry import DeviceInfo
 from homeassistant.helpers.typing import ConfigType
@@ -135,6 +135,13 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
             control_board,
         )
         control_board = None
+        try:
+            await hass.async_add_executor_job(grills.get_grill, model, None)
+        except InvalidGrill as ex:
+            # Not "not ready": `api.start()` would raise this again on every
+            # retry, and no amount of waiting introduces a model to the
+            # catalogue. Asking the user to reconfigure is the only way out.
+            raise ConfigEntryError(f"Unknown grill model: {model}") from ex
 
     # Constructed inline: since spec resolution moved into `start()`, the
     # constructor only assigns attributes. The executor job it ran in was a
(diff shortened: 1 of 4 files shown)


dknowles2 (approved):
Both are the same shape of bug and it is a good one to have gone looking for: a permanent failure wearing a transient failure's clothes. Retrying is not neutral when it cannot work — it buries the one message that would let someone fix it.

**The unknown-model path checks out.** `_start_api` catches only `GrillUnavailable`:

```python
try:
    await self.api.start()
    self._api_started = True
except GrillUnavailable as ex:
    raise UpdateFailed("Grill unavailable") from ex
```

so an `InvalidGrill` from `get_grill` escapes as an unexpected error and becomes `ConfigEntryNotReady` — an escalating backoff for a model that will never appear in the catalogue by waiting. Validating the fallback rather than trusting it closes the hole #354 left open: that PR checked the *prefix* pairing and dropped the board on failure, but nothing checked the model itself resolved either way.

`ConfigEntryError` rather than `ConfigEntryNotReady` is the right distinction and the one Home Assistant actually acts on differently.

**The password case is the better of the two**, because the failing user is already in the worst position: they know the password is wrong, they are trying to fix it, and the service hands them a flat toast. Starting the reauth flow gives them the same route the coordinator would have offered — and doing it from the service means they get there without waiting for a poll to fail.

One detail worth noting since it has bitten this repo before: `except Unauthorized` sits above `except Exception`, and it has to — `Unauthorized` subclasses `RPCError` subclasses `Exception`, so the order is load-bearing exactly as it was in #350. Leaving the entry alone and only starting the flow is also right; nothing here knows the new password is good.

CI green: build, mypy, ruff, Hassfest, HACS. **I could not run the tests** — the suite does not start here — so CI is the check and I did not reproduce either mutation.

Labelled `bug`. Merging.

PR #366, opened 2026-08-03
Back off, and say why, when a cycle fails

trailro:
A grill that is switched off while its connection stays up — the normal case on the cloud relay, where the socket outlives the grill — keeps polling at the active cadence and reports the wrong reason for it. Two halves of the same cycle.

## The interval never backs off after a failure

`_apply_poll_interval` runs only after a successful `get_state`. Any earlier failure returns first, so `update_interval` stays at whatever the last **successful** read set.

So: grill lit → `ACTIVE_SCAN_INTERVAL` (10 s). User switches it off. `api.is_connected()` is still true, because the relay socket is still open — the `ConnectivitySensor` docstring already describes exactly this. Every cycle now reaches `api.ping(timeout=10.0)` and burns the full ten seconds timing out, and the interval never reaches `STANDBY_SCAN_INTERVAL`. A ten-second ping permanently in flight for as long as the grill is off, which is the opposite of what the two constants exist for.

Backing off on failure is now in one place: `_async_update_data` wraps the cycle, and the body moves to `_async_poll`.

## `TimeoutError` is not caught, and it is the likely one

`Transport.send_command` raises `TimeoutError` when a reply does not arrive — which is precisely what happens above. Only `NotConnectedError` was caught, at both the `ping` and the `get_state` call sites, so the timeout reached Home Assistant's generic handler and the entry's failure reason became *"Timeout fetching pitboss data"* instead of the intended `UpdateFailed`.

Both sites now treat it as a failed cycle with a reason that says what happened.

## Testing

Two tests, both **verified to fail on the unpatched code**: a ping timeout surfaces as `UpdateFailed`, and a failed cycle leaves the interval at standby rather than active.

133 pass, `ruff check`, `ruff format --check` and `mypy .` clean, run in a Linux container as on #360, #362 and #363 — the suite does not start on macOS, where setting up the `bluetooth` component dies in `dbus_fast`.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -286,6 +286,20 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
             self.logger.debug("Could not fetch the system info: %s", ex)
 
     async def _async_update_data(self) -> StateDict:
+        try:
+            return await self._async_poll()
+        except UpdateFailed:
+            # Back off before giving up on this cycle. The interval is
+            # otherwise whatever the last *successful* read set, so a grill
+            # that goes off while the connection stays up -- which is the
+            # normal case on the cloud relay, where the socket outlives the
+            # grill -- keeps a ten-second ping in flight for as long as it
+            # is off, which is the opposite of what the two intervals exist
+            # for.
+            self._apply_poll_interval(StateDict())
+            raise
+
+    async def _async_poll(self) -> StateDict:
         if not self._api_started:
             self.logger.debug("Starting API")
             await self._start_api()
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both halves check out, and the second is the one I would not have found by reading the diff.

**`TimeoutError` really is uncaught, and really is the likely one.** `send_command` wraps the exchange in `asyncio.timeout(timeout)`, so a reply that never arrives raises `TimeoutError` — not `NotConnectedError`, which was the only thing either call site caught. So the failure that a switched-off grill behind a live relay socket produces *by construction* was the one falling through to Home Assistant's generic handler.

**And the interaction between the two halves is what makes this worth doing as one change.** `_apply_poll_interval` ran only after a successful `get_state`, so the interval stayed at whatever the last success set — meaning a grill switched off mid-cook holds `ACTIVE_SCAN_INTERVAL` and spends ten of every ten seconds inside a ping that will time out. That is not just failing to back off; it is a ten-second timeout permanently in flight, which is precisely the load #351's two constants exist to avoid.

Moving the body to `_async_poll` and backing off in the wrapper puts it in one place rather than at each early return, which is what let it be missed in the first place.

## One tradeoff worth recording

Backing off on *any* failed cycle means a single transient hiccup during a cook drops to `STANDBY_SCAN_INTERVAL`, so the next reading is up to 60 seconds away rather than 10 — a blind spot at exactly the time the user is watching.

I think you have it the right way round: the case being fixed is permanent and costly, the case being traded is momentary and self-correcting on the next success. Backing off from a link that just failed is also the better instinct than hammering it. But if stale-during-cook reports ever appear, this is the line, and "after N consecutive failures" would be the knob.

Not worth adding now — a counter for a case nobody has reported would be inventing state.

CI green: build, mypy, ruff, Hassfest, HACS. **I could not run the tests** — the suite does not start here — so CI is the check and I did not reproduce either of your mutations. The two new tests land in an existing `test_coordinator.py` alongside twelve others, which is where they belong.

Labelled `bug`. Merging.

PR #368, opened 2026-08-04
Bring the README up to date, and make that part of the work

dknowles2:
Docs only. No code changes.

## What was stale

The README described the integration as it was a release ago. It listed **six platforms where eight are registered**, said probe targets existed for **probes 1 and 2** when #341 gave every probe one, and never mentioned buttons, the grill-unit select, the connectivity sensor, target-reached sensors, the meat probe control sensor, firmware and controller diagnostics, adaptive polling, reauth, or the `set_grill_password` action.

## What was wrong

Different problem, and worth separating out. The Features list opened with:

> **100% local, no cloud required.** Communicates directly with the grill over Bluetooth (BLE) or your local WiFi (WebSocket)

`wss` is not local WiFi. It connects outbound to `socket.dansonscorp.com` and the vendor relays to the grill — and `DEFAULT_PROTOCOL = PROTOCOL_WSS`, so that is what most installs are doing. The page told users the opposite of what the default does, in bold, as its headline feature. "No cloud access necessary!" appeared again lower down.

There is now a table saying plainly which protocol reaches the grill how and which needs internet, with `ble` named as the option for people who want no cloud in the path. The genuinely local transport is described as what it is: coming to pytboss, not wired up here, and not available on every grill.

## Checked against the code, not the release notes

- platform list ← `PLATFORMS` in `__init__.py` (8, and the table now has 8)
- entity names ← `_attr_name` across `custom_components/pitboss/`
- probe labelling ← `probe_label()`, which follows the vendor app: `MPC` only on grills with a control port, and the numbering does not shift
- "request fast updates" ← `button.py`, which says it does nothing unless the grill is on and nothing off the cloud path — so the README says that too, rather than "a burst of frequent updates"
- setpoint snapping, held probe targets, switches staying available ← `coordinator.py`

## The AGENTS.md half

Adding the missing entities fixes today. The reason it drifted is that nothing asks for the README when an entity is added, and nothing fails when it is skipped — **every stale claim above was correct when it was written**.

So `AGENTS.md` now names updating `README.md` as part of the change rather than follow-up, lists the five kinds of change that require it, and gives the two greps that answer "did I miss an entity":

```sh
grep -A 12 'PLATFORMS: list\[Platform\]' custom_components/pitboss/__init__.py
grep -rn '_attr_name' custom_components/pitboss/
```

It also records the three specific drifts above, since a rule with the failure attached tends to survive better than a rule on its own.

## Verified

`ruff check .` and `mypy .` clean.

**The test suite is not a signal here and I am not presenting it as one** — it fails identically on unmodified `main` (116 failed, 17 passed, `KeyError: 'pitboss'`). This changes no code, so CI is the check.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,45 @@ merge in order — a PR that reads a new attribute is blocked on the release
 *and* on the bump landing first, and neither CI nor GitHub can express that.
 Say so in the PR.
 
+## Update the docs in the PR that changes the behaviour
+
+`README.md` is what users read to decide whether this integration does what
+they want. It is part of the change, not follow-up work — a PR that adds an
+entity and leaves the README alone is incomplete, and the gap is invisible in
+review because nothing fails.
+
+Update it in the same PR when a change:
+
+- **adds, removes or renames an entity**, or changes when one is created —
+  the platform table and the entity list both name specific entities;
+- **adds a platform** — the table lists platforms, and `PLATFORMS` in
+  `__init__.py` is the list to check it against;
+- **adds or changes an action** — see the Actions section;
+- **changes what a user sees or has to do**: setup and reconfigure steps,
+  repair flows, a new option, polling behaviour they would notice;
+- **changes something the README states as a fact** — connection protocols,
+  what is and is not local, or the safety position on remote start.
+
+Two things to check rather than assume:
+
+```sh
+# Every platform the integration actually sets up.
+grep -A 12 'PLATFORMS: list\[Platform\]' custom_components/pitboss/__init__.py
+# Every entity name a user will see.
+grep -rn '_attr_name' custom_components/pitboss/
+```
+
+The README drifted a whole release behind once — it described probe targets
+as existing for probes 1 and 2 only, listed six platforms when eight were
+registered, and called the `wss` protocol "local WiFi" when it connects
+outbound to `socket.dansonscorp.com`. Each of those was correct when written,
+and none of them failed a test.
+
+`CONTRIBUTING.md` and `REVIEW.md` change far less often; touch them when the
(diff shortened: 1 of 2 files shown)


dknowles2:
Added the connection rename @dknowles2 asked for, in this PR rather than a new one — the README describes those names, so splitting them would ship the docs describing labels that did not exist yet. (Which is what the AGENTS.md rule this PR adds asks for.)

## What changed

The picker offered **"Bluetooth (local)"** and **"WebSocket (cloud)"**. It now offers **"Bluetooth"** and **"Cloud"**.

"WebSocket" is a transport protocol, not a thing anyone is choosing between — the parenthetical was doing all the work, so it is now the label.

The field is **"Connection"** rather than "Connection protocol", and its hint says what the choice does instead of restating the field name:

> ~~The protocol to use to connect to your grill.~~
> Bluetooth talks to the grill directly. Cloud connects through PitBoss's servers, so it needs internet, but it is not limited by Bluetooth range.

That is the trade someone is actually making, and the second half matters — Bluetooth range is why the cloud option is the default.

## Labels only

**The stored values stay `ble` and `wss`.** `PROTOCOL_BLE` and `PROTOCOL_WSS` are untouched, so no config entry changes and nothing needs migrating. The selector already went through `translation_key="protocol"`, so the whole change is `translations/en.json` — six lines.

The README uses the new names and notes that diagnostics and stored configuration still show `ble`/`wss`, so someone reading a diagnostics dump can tell it is the same thing.

`pt.json` has no `selector` section, so Portuguese falls back to English as it already did. I have not added a translation I cannot check.

## Verified

`ruff check .` and `mypy .` clean. Still no code change beyond the JSON, so CI remains the check for the suite.

PR #369, opened 2026-08-04
Call the connection choice Bluetooth and Cloud

dknowles2:
Re-raising the connection rename. It was pushed to #368 but that PR merged at the commit before it, so the change never landed — `main` still shows the old labels. Same commit, no changes.

## What changes

The picker offers **"Bluetooth (local)"** and **"WebSocket (cloud)"**. It will offer **"Bluetooth"** and **"Cloud"**.

"WebSocket" is a transport protocol, not a thing anyone is choosing between — the parenthetical was doing all the work, so it becomes the label.

The field is **"Connection"** rather than "Connection protocol", and its hint says what the choice does instead of restating the field name:

> ~~The protocol to use to connect to your grill.~~
> Bluetooth talks to the grill directly. Cloud connects through PitBoss's servers, so it needs internet, but it is not limited by Bluetooth range.

That second half is the trade someone is actually making, and it is why the cloud option is the default.

## Labels only

**The stored values stay `ble` and `wss`.** `PROTOCOL_BLE` and `PROTOCOL_WSS` are untouched, so no config entry changes and nothing needs migrating. The selector already went through `translation_key="protocol"`, so the whole functional change is `translations/en.json` — six lines.

The README section is renamed to match and notes that diagnostics and stored configuration still show `ble`/`wss`, so someone reading a diagnostics dump can tell it is the same thing.

`pt.json` has no `selector` section, so Portuguese falls back to English as it already did. I have not added a translation I cannot check.

## Verified

```
labels:          {'ble': 'Bluetooth', 'wss': 'Cloud'}
keys unchanged:  True
```

`ruff check .` and `mypy .` clean. No Python changes, so CI is the check for the suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,22 +23,26 @@ dependency is updated.
 
 - **No account, no app.** The integration talks to the grill itself. It never signs in to a PitBoss account and never uses the vendor's mobile app.
 - **Automatic discovery.** Power on your grill and it's discovered automatically over Bluetooth; manual setup by device ID is also supported.
-- **Two connection protocols.** Bluetooth (`ble`) or WebSocket (`wss`, the default), chosen at setup and changeable later via reconfigure. See [Connection protocols](#connection-protocols) — they are not equivalent, and one of them is not local.
+- **Two ways to connect.** **Bluetooth** or **Cloud** (the default), chosen at setup and changeable later via reconfigure. See [How it connects](#how-it-connects) — they are not equivalent, and one of them is not local.
 - **Reconfigurable, and it asks rather than gives up.** Change the model, password, or protocol without deleting the integration. If the grill starts rejecting the password, the integration asks you to re-enter it instead of retrying forever.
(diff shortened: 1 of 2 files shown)


dknowles2:
Closing this — @trailro adopted the three labels directly in #371 (`0889639`):

```json
"ble": "Bluetooth",  "wss": "Cloud",  "local": "Local"
```

which is what this PR existed for, plus the third option it could not have known about. Landing both would just be a conflict over the same JSON object.

The rest of this PR is superseded too: the README section it edited was rewritten by #368 and again by #371, which now carries its own Connection protocols section.

**What is genuinely still undone**, and not worth reviving this stale branch for:

- the field label is still "Connection protocol" rather than "Connection"
- its hint is still "The protocol to use to connect to your grill.", which restates the field name instead of saying what the choice does

Both are worth fixing, but the hint I wrote here names only Bluetooth and Cloud, so it needs rewriting for three options regardless. Cleaner as a fresh two-line change once #371 lands than as a rebase of a branch that is now ten merges behind.

Happy to raise that follow-up whenever #371 is in.

PR #370, opened 2026-08-04
Add a start_grill action behind an opt-in

### LINKED ISSUE #322: Remote power-on is possible: FE0101FF (verified on PBL)
Split out from #321 for tracking, as requested.

### What

The control boards accept `FE0101FF` as a power-on command — the mirror of the `FE0102FF` we already send for off, and the same shape as the light pair (`FE0201FF` / `FE0202FF`):

```lua
function pitboss_api.set_power(ip_address, state)
	local command = (state == "on" or state == true) and "FE0101FF" or "FE0102FF"
	return pitboss_api.send_command(ip_address, command)
end
```

pytboss declares no `turn-on` slug for any board, so it has to be sent as a raw hex command.

### Verified

On a **PB1600PS1 (PBL board)**, over the local WiFi transport:

- starting state: off, cold (25 °C), no errors, auger and fan idle
- `PB.SendMCUCommand` with `FE0101FF` → `{"result": null}`
- the grill lit; I shut it down immediately from the panel

I did this standing at the grill with the lid open and a clear burn pot. Worth repeating that precaution if you test it: the documented hazard when starting a pellet grill is unburned pellets accumulating in the pot and deflagrating, which is why the manuals all say to start with the lid open.

### Why it matters

The comments in `climate.py` and `switch.py` say the grill *cannot* be turned on remotely. It can — the restraint is a deliberate choice, which is a different (and defensible) thing. Even just rewording those comments would stop the next person assuming the capability is absent, as I did.

### If you want it exposed

My suggestion is a dedicated service, never `switch.turn_on` and never `HVACMode.HEAT` — those are the surfaces that fire from a stray tap or a mis-scoped automation. For what it's worth, in my own fork I gate it three ways: an options-flow toggle that defaults to off, a refusal if the grill is already on or disconnected, and a refusal on any of `noPellets`, `highTempErr`, `erL`, `fanErr`, `motorErr`, `hotErr` — naming which one blocked it.

Happy to contribute that, or to leave it entirely alone. It's your call and I don't think a contributor should make it for everyone.

trailro:
Closes #322. Needs `pytboss 2026.8.4`, which `main` already pins.

Adds `pitboss.start_grill`, taking a device and nothing else. It calls `api.turn_grill_on()` — the method you asked for on #322 before anything downstream used it, added in [pytboss#550](https://github.com/dknowles2/pytboss/pull/550).

**Lighting a grill remotely is the most consequential thing this integration can do**, so the gating is the substance of the PR rather than a wrapper around it:

- **Off by default.** Refuses unless `enable_remote_start` is switched on in the integration options.
- Refuses if the grill is already on, or not connected.
- Refuses on `noPellets`, `highTempErr`, `erL`, `fanErr`, `motorErr`, `hotErr`. The board would likely fail anyway; failing here says which one.
- Logs a warning naming the lid and the burn pot before sending.
- Takes a **device**, so it does the right thing on a multi-grill setup rather than guessing at an entry.

There is a test per error flag rather than a sample of them, so removing one from the dict fails exactly its own test.

## It introduces the first options flow

The toggle has to live somewhere and this repo has none. So this adds one, with a single field, built so the next thing that wants an option extends it rather than bringing a second.

Two things about that:

**I recommended closing [#277](https://github.com/dknowles2/ha-pitboss/pull/277)** — its approach cannot work, because Home Assistant re-derives an entity's displayed unit from the unit system on every update, so an integration-level override never reaches the user. I demonstrated it and corrected my own earlier explanation of the mechanism there. If you would rather keep that PR and have it own the flow, say so and I will rebase onto it and drop the flow from here.

**It carries the update listener** whose absence I flagged on #277. Not needed for this toggle, which is read live — but an options flow without one is a trap for the next option added, and I would rather not build the trap.

## On testing

**Confirmed on hardware by both of us** — you on your PBV4PS2 as you said on the issue, me on a PB1600PS1 with the lid open and the burn pot clear.

**I have not re-run it since, deliberately.** There is no version of "just checking" this that is safe to do with nobody standing next to the grill, so everything here is verified against the released library and the test suite, and the live behaviour rests on those two runs.

142 tests, ruff, `ruff format --check` and mypy clean.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -168,7 +168,19 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
         raise
 
     await hass.config_entries.async_forward_entry_setups(entry, PLATFORMS)
+    entry.async_on_unload(entry.add_update_listener(_async_options_updated))
     return True
+
+
+async def _async_options_updated(hass: HomeAssistant, entry: ConfigEntry) -> None:
+    """Reload when the options change.
+
+    The remote-start toggle is read live, so this is not strictly needed for
+    it -- but an options flow without a listener is a trap for the next
+    option added, and Home Assistant fixes some entity properties (a display
+    unit, for one) when the entity is registered.
+    """
+    await hass.config_entries.async_reload(entry.entry_id)
 
 
 async def async_unload_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -129,7 +129,8 @@ class GrillClimate(BaseEntity, ClimateEntity):
             await self.async_turn_off()
         elif hvac_mode == HVACMode.HEAT:
             LOGGER.warning(
-                "For safety reasons, the grill cannot be turned on remotely."
+                "Not lighting the grill from here. Use the pitboss.start_grill "
+                "action, which has to be enabled in the integration options first."
             )
 
     @property
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -7,7 +7,12 @@ from typing import Any
 
 import voluptuous as vol
 from homeassistant.components.bluetooth import BluetoothServiceInfoBleak
-from homeassistant.config_entries import ConfigFlow, ConfigFlowResult
+from homeassistant.config_entries import (
+    ConfigEntry,
+    ConfigFlow,
+    ConfigFlowResult,
+    OptionsFlow,
+)
 from homeassistant.const import CONF_DEVICE_ID, CONF_MODEL, CONF_PASSWORD, CONF_PROTOCOL
 from homeassistant.helpers.selector import SelectSelector, SelectSelectorConfig
 from pytboss import grills
@@ -25,6 +30,11 @@ def _models_on_board(control_board: str) -> list[str]:
 
 class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
     """Config flow for PitBoss."""
+
+    @staticmethod
+    def async_get_options_flow(config_entry: ConfigEntry) -> PitBossOptionsFlow:
+        """Return the options flow handler."""
+        return PitBossOptionsFlow()
 
     VERSION = 1
     MINOR_VERSION = 3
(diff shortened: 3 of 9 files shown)


trailro:
Re-read the issue against this PR and found one thing I had dropped: **the misleading comments.** Fixed at `45642cb`.

The issue body asks for it directly, and my own comment called it "worth doing either way":

> The comments in `climate.py` and `switch.py` say the grill *cannot* be turned on remotely. It can — the restraint is a deliberate choice, which is a different (and defensible) thing.

Both said:

> For safety reasons, the grill cannot be turned on remotely.

Which was arguably true when written and is plainly false the moment this PR lands. They now say:

> Not lighting the grill from here. Use the pitboss.start_grill action, which has to be enabled in the integration options first.

That is the difference between a capability the integration lacks and a surface it deliberately declines to expose — and the wrong version is what made me assume the capability was absent in the first place, which is how this issue started.

The two surfaces still refuse, unchanged: `switch.turn_on` and `HVACMode.HEAT` are exactly the ones that fire from a stray dashboard tap or a mis-scoped automation, and neither should light a fire.

### Against the rest of the issue

Everything else you asked for is here, and nothing in it has changed shape since you approved the design:

- a dedicated action, never `switch.turn_on` and never `HVACMode.HEAT`
- an options toggle defaulting to off
- refusals on already-on, disconnected, and each of `noPellets`, `highTempErr`, `erL`, `fanErr`, `motorErr`, `hotErr`, naming which one blocked it
- a warning on every use, naming the lid and the burn pot
- the pytboss method first, which is [#550](https://github.com/dknowles2/pytboss/pull/550)

142 tests, ruff, `ruff format --check` and mypy clean.

dknowles2 (approved):
Verified the field names against `pytboss.grills.StateDict` (`moduleIsOn`, `noPellets`, `highTempErr`, `erL`, `fanErr`, `motorErr`, `hotErr` all real) and that `turn_grill_on()` exists on `main`'s pinned pytboss (#550, released in 2026.8.4). The gating is exactly as thorough as the description claims — off by default, per-error test coverage that would catch a dropped flag, device-scoped lookup.

One gap: no test exercises the `is_connected() is False` refusal path specifically (only the enabled/already-on/error branches are). Small, not blocking given how the rest is covered.

Options-flow-ownership question for #277 is David's call, not mine.

PR #371, opened 2026-08-04
Add a local WiFi connection option

trailro:
The Home Assistant half of item 16 in #344. The transport itself is your [pytboss#543](https://github.com/dknowles2/pytboss/pull/543), released in `2026.8.4`, which `main` already pins.

Adds a third protocol option, `Local (WiFi)`, with a host field and the `HttpConnection` branch in `async_setup_entry`. That removes the Dansons relay from the path entirely for grills that can take it.

## The address is validated by trying it

**Not every grill has the local endpoint**, and I said on #505 and #344 that the flow should probe `Config.Get` for `http.enable`. It attempts the connection instead, and I think that is the better answer:

- At the point the user picks a protocol there is **no other transport to ask over**. That suggestion assumed a grill already connected, which is only true for reconfigure.
- `http.enable` predicts whether anything is listening. `HttpConnection.connect()` establishes that something *is*, and catches a typo'd address with the same error.
- It also covers the second firmware family from [#558](https://github.com/dknowles2/pytboss/pull/558), where there is no HTTP server at all and no `http` section in the config to read.

A grill with the endpoint off and a wrong IP are the same thing from the user's side, and the message says so rather than pretending to know which:

> No grill answered at that address. Not every grill has the local interface enabled, and it cannot be turned on from here.

Nothing here writes config to turn the endpoint on, per the conclusion on #505.

Validation runs in both `more_info` and `reconfigure`, so switching an existing grill over is checked the same way as setting one up.

## Verified against the released package

**137 tests, ruff, `ruff format --check` and mypy clean** against `2026.8.4` itself rather than a branch overlay. Four new tests: host required, an unreachable host reported rather than accepted, the entry created with the host stored, and other protocols unaffected by the field.

Running against a real grill too — PB1600PS1, firmware 0.5.7, `http.enable` true — which is where I found the two defects fixed in [#546](https://github.com/dknowles2/pytboss/pull/546).

## Two notes

**Every entry now carries a `host` key**, empty unless the local protocol is chosen. `test_user_flow_creates_entry` asserted the exact data dict and now includes it — flagging that rather than letting an edited assertion pass unnoticed.

**This conflicts textually with [#370](https://github.com/dknowles2/ha-pitboss/pull/370)** — both touch `config_flow.py`, one adding a host field and the other an options flow. No shared logic; whichever lands second I will rebase.

## Still open, and not in this PR

The host has to be typed. Two routes to changing that, neither built: firmware 0.6.0 broadcasts `{currentIP: "..."}` over Bluetooth and virtual data, so a paired grill could pre-fill it; and the device id is MAC-derived, so a `dhcp` matcher may be possible — untested, and I would not claim it works. Neither helps a 0.5.7 grill that has never been paired, so the field has to exist regardless.


trailro:
Adapted to the contract `2026.8.4` actually ships, rather than only rebasing onto it — pushed at `1eae894`.

`_validate_local` caught bare `Exception` when I wrote it, because before [#571](https://github.com/dknowles2/pytboss/pull/571) `connect()` could surface a `JSONDecodeError` that nothing in the contract named. That PR made the outcomes precise, so collapsing them into one message is now throwing away information the library is going to the trouble of giving us:

| what happened | now says |
|---|---|
| `NotConnectedError` — nothing listening | No grill answered at that address |
| `Unauthorized` — 401 at the HTTP layer | A grill answered but refused the connection |
| `RPCError` — answered, but not JSON-RPC | Something answered, but it is not a grill |

The third is the one worth having. A mistyped IP landing on a router's admin page answers `200 text/html`, which is exactly the mistake a user makes, and "no grill answered" would send them looking for the wrong problem. That case only became distinguishable because of #571.

The second is rare — only a Mongoose build with `rpc.auth_file` refuses at the HTTP layer rather than in a JSON-RPC error frame — but it costs nothing to be right about, and "no grill answered" would be wrong rather than vague.

A test per branch, parametrised, and I verified that collapsing them back to one message fails exactly the two that should.

**140 tests, ruff, `ruff format --check` and mypy clean** against the released `2026.8.4`, on current `main` at `f10dcad`.



---

*Edited: I wrote `d3aeafc` above; the commit is `1eae894`. Third time I have put a hash in a comment that I composed before running the push — the fix is ordering, not care, so from here the comment gets written after the push and the hash is pasted from its output.*

trailro:
Pushed one more commit (`d87f3d0`) closing a gap I found reviewing my own validator, before it found a user.

**A silent host could crash the flow instead of saying "no grill answered."** `_validate_local` caught `NotConnectedError`, `Unauthorized` and `RPCError` — but on an address where nothing answers, the transport races its own `asyncio.timeout` against aiohttp's `ClientTimeout`, both at the 30s default, and when its own deadline fires first the failure surfaces as a **bare `TimeoutError`**. Uncaught, that reached Home Assistant's flow handler as "Unknown error occurred" — the vaguest possible answer on the most common failure path this validator exists for. It now maps to `cannot_connect` alongside `NotConnectedError`, with a comment saying why it is caught in its own right; to the user both are "nothing answered". pytboss#576 straightens the race out at the transport, and catching both here stays correct whichever deadline wins, before and after that lands.

**Validation also stops borrowing the transport's 30s deadline.** It passes its own `timeout=10.0` — this holds a form open while the user waits, and a silent address is the very failure it exists to catch quickly.

One more case in the parametrized failure test covers the bare timeout. 141 tests, ruff, `ruff format --check`, mypy clean; CI green on the updated branch.

dknowles2 (dismissed):
Blocking: \`config_flow.py\` line 58 does not parse.

\`\`\`python
    except NotConnectedError, TimeoutError:
\`\`\`

That's Python 2 syntax — needs to be \`except (NotConnectedError, TimeoutError):\`. Confirmed with \`ast.parse\` against both a plain \`git fetch\` of the PR head and a direct GitHub Contents API fetch of blob \`d87f3d0\`, under both Python 3.9 and 3.12:

\`\`\`
SyntaxError: multiple exception types must be parenthesized
\`\`\`

This means \`custom_components/pitboss/config_flow.py\` cannot be imported at all — since Home Assistant loads this module to set up the config flow, this would break setup for every protocol, not just the new local one.

Oddly, the CI run for this exact head SHA (https://github.com/dknowles2/ha-pitboss/actions/runs/30890967213) reports "141 passed" including \`test_local_protocol_checks_the_grill_answers\`, which shouldn't be possible if the module can't import. I haven't chased down why — possibly a checkout-ref/caching quirk in that run — but the syntax error itself isn't in question; I verified it against the raw blob independent of CI. Worth a look regardless of this PR, since it means green CI isn't currently a fully reliable signal on this branch.

Rest of the PR looks right on read — the validation-by-connecting approach, the per-error-type messaging, the host field threaded through both \`more_info\` and \`reconfigure\`. Will re-review once this line is fixed to \`except (NotConnectedError, TimeoutError):\`.

trailro:
Resolved against #370, as promised when both were opened — `main` merged in (append-only, as with #356 and #374), and the diff is back to the local-transport change alone.

The conflicts were the three predicted compositions in `config_flow.py`, all unions with nothing to choose between: the import blocks (its `OptionsFlow` imports beside this PR's `CONF_HOST`), the `.const` import list, and the file tail where its `PitBossOptionsFlow` class now sits after this PR's `reconfigure` return. `const.py`, `translations/en.json` and `__init__.py` merged themselves.

Full suite green on the merged branch — 152 passed, both features' tests side by side, in the usual `python:3.14` container — with ruff, `ruff format --check` and mypy clean.

trailro:
Two more commits from a full pass over this PR's angles before it reaches you — one real defect, two completeness gaps.

## The defect: one network blip stranded the local protocol until a reload (`9a56501`)

`HttpConnection` marks itself disconnected when a request fails and, by its own design, recovers on the next request — `_send_prepared_command` deliberately gates on `_started`, not `_connected`. But the coordinator refuses to speak to a transport reporting disconnected (`if not self.api.is_connected(): raise UpdateFailed`), and unlike the other two protocols nothing else ever speaks for it: Bluetooth is reconnected by the discovery callback, the websocket by its internal loop, HTTP by **nobody**. Sequence: grill drops off the network for one poll → `_connected` goes false → every later cycle short-circuits at the gate → the transport's only recovery path is never taken. Permanent `Grill not connected` on the transport that exists to be the reliable one.

Fix: the coordinator takes `reconnect_on_poll` (set only for the local protocol) and, when disconnected, re-runs `_start_api()` instead of failing the cycle outright — `connect()` is the transport's own reachability probe, so success means connected again, and a probe that fails maps `NotConnectedError` to `UpdateFailed` like every other miss. **Deliberately not generic:** `connect()` on a websocket mid-backoff starts a second receive loop on the pinned `2026.8.4` — that lifecycle was serialized upstream in pytboss#572, merged but not yet released, so the opt-in stays until a release carries it. Three new coordinator tests: the blip-then-recovery cycle, a failed reconnect degrading to a failed cycle, and the default still failing fast with `start` never awaited.

## The completeness gaps

**README (`e5a77cb`):** it still said the local option "is not wired up here yet" — false the moment this merges, and #368 made the README part of the work. Third row in the protocol table, a paragraph on which grills actually have the endpoint (the ESP-IDF `16.x` line has none; Mongoose units need `http.enable`) and that the form verifies by connecting, and the manual-setup section now mentions the host field.

**Setup-path test (`9a56501`):** `test_init` covered the BLE and WSS branches of `async_setup_entry` but not the new one. Now mirrored: local entry loads, `HttpConnection` is built with the host and the shared Home Assistant session — the one `disconnect()` must not close.

Angles checked and left alone, so the review does not have to re-derive them: `Sys.Reboot` over HTTP is safe on `2026.8.4` (background task, held in `_pending`, cancelled on disconnect); the fast-updates button correctly does not exist on this protocol; the validator's owned session closes on every path including the failed-connect double-disconnect; reauth keeps the host and a runtime 401 still lands in the reauth flow; first-time setup against an unreachable grill still retries as `ConfigEntryNotReady`.

156 tests, ruff, `ruff format --check`, mypy clean; suite run against released `2026.8.4` in the usual `python:3.14` container.

dknowles2:
Two more commits landed (`9a56501`, `e5a77cb`) since the blocking review, but line 64 of `config_flow.py` is unchanged — still `except NotConnectedError, TimeoutError:`. Module still won't import. Review stays CHANGES_REQUESTED until that's `except (NotConnectedError, TimeoutError):`.

trailro:
Fixed at `dcca03b`, and your CI mystery has a mundane answer that is worth having on record.

**Why green CI "shouldn't be possible" but was:** `.python-version` pins CI to **3.14**, and on 3.14 `except A, B:` is *valid* — PEP 758 made unparenthesized multiple exception types legal again. Your `ast.parse` under 3.9 and 3.12 rejects it; CI's 3.14 parses it happily and ran all 141 tests for real. No checkout quirk, no caching — a version-floor mismatch between what CI runs and what users run, since `hacs.json` still allows Home Assistant back to 2024.1 and those installs are on older interpreters. Your instinct that the module had to fail to import was right *for the floor*, which is exactly why this needed fixing.

**Why the fix is two handlers rather than the parentheses you asked for:** I wrote the parenthesized tuple originally — `ruff format`, targeting `py314`, **rewrote it into the unparenthesized form**. That is how the line got this way, and it would do it again on the next format pass; with the parentheses restored, `ruff format --check` fails the build. So the commit catches `NotConnectedError` and `TimeoutError` in two separate handlers saying the same thing — syntax every interpreter parses and the formatter has no opinion on — with a comment explaining why they are not a tuple, so nobody helpfully refolds them.

**The repo-level decision that leaves open, flagged rather than made here:** as long as `target-version = "py314"` in `.ruff.toml` sits above the interpreter floor implied by `hacs.json`, the formatter will keep being willing to rewrite code into 3.14-only syntax that CI then validates. Either the ruff target drops to the true floor (py312, so the formatter physically cannot produce this again) or the HACS minimum rises to match — your call, and happy to send either as a follow-up.

Also merged `main` in — this had gone BEHIND when #374 landed; no conflicts. 158 tests (the suite grew with #374's), ruff, `ruff format --check`, mypy clean, and the fixed module now parses under `ast.parse(..., feature_version=(3, 12))` as well as 3.14.

dknowles2:
Confirmed at `dcca03b`: `config_flow.py` parses under 3.9 and 3.12 now — two separate `except` clauses instead of the tuple form, avoiding the PEP 758 rewrite. Your explanation for the earlier "green CI is impossible" mystery (`.python-version` pinned to 3.14, which accepts unparenthesized multi-exception `except` natively) matches what's in the repo.

Read through the two new commits since the blocking review (`9a56501`, `e5a77cb`) as well as the fix itself:

- `reconnect_on_poll` in the coordinator looks right for the stated problem — HTTP has no background reconnect, so the poll loop has to be the one to re-issue `connect()`. The three new coordinator tests cover: reconnect succeeds, reconnect fails with `NotConnectedError`, and the opt-out transports still fail fast without calling `start()`.
- One thing not covered, non-blocking: `_start_api()` only catches `GrillUnavailable` and `NotConnectedError`. `HttpConnection.connect()` can also raise `Unauthorized` (a build with `rpc.auth_file` set) or `RPCError` (a non-JSON or non-dict response) — both reachable, per the handling `_validate_local` already gives them in this same PR — and neither is caught here. On the reconnect-on-poll path that would surface as an unhandled exception out of `_async_update_data` instead of a clean `UpdateFailed`, where every other failure in this method lands. Worth a look, but doesn't need to hold up the rest of this.

Nothing else outstanding from my side. Not merging — this now has an unresolved CHANGES_REQUESTED on file from before the fix, and I don't clear that myself.

trailro:
The non-blocking gap is closed too, at `80bb923`: `_start_api` now maps `Unauthorized` and `RPCError` from the reconnect probe to `UpdateFailed`, where every other failed cycle lands. Deliberately *not* `ConfigEntryAuthFailed` for the `Unauthorized` case — an `rpc.auth_file` refusal is HTTP-layer auth, not the grill password, so opening the reauth flow would send the user to change the one credential that cannot fix it. The failed-reconnect test is now parametrized over all three probe outcomes.

One wrinkle worth a smile: the new tuple is safe from the formatter because it has `as ex` — PEP 758 legalized `except A, B:` but still requires parentheses with `as`, so this is the one spelling ruff cannot rewrite.

160 tests, ruff, `ruff format --check`, mypy clean. Re-requesting review to clear the CHANGES_REQUESTED on file.

dknowles2 (commented):
Not clearing @dknowles2's `CHANGES_REQUESTED` — that is his to lift. But the blocking line is fixed at the current head, and the part he flagged as unexplained has an explanation that matters beyond this PR.

## The syntax error is gone, and CI was not lying

`80bb923` parses cleanly:

```
feature_version 3.9:  parses OK
feature_version 3.12: parses OK
```

@trailro split the handler in two rather than parenthesising it, and the comment explaining why is the important part.

**CI reported "141 passed" honestly.** It was not a checkout-ref or caching quirk. Every workflow here uses `python-version-file: '.python-version'`, and that file says `3.14`. `except NotConnectedError, TimeoutError:` is **valid syntax on 3.14** — it is PEP 758, new in that release. @dknowles2 parsed it under 3.9 and 3.12, where it is not. Both observations were right about different interpreters.

## The suggested fix would have put the bug back

`except (NotConnectedError, TimeoutError):` does not survive `ruff format` in this repo. Reproduced against the repo's own config:

```
before:  except (ValueError, TypeError):
after:   except ValueError, TypeError:
         -> 3.12 SyntaxError: invalid syntax
```

There is no `ruff.toml` and no `[tool.ruff]` in `pyproject.toml`, so ruff infers its target from `.python-version` — `3.14` — and rewrites the parenthesised form into the newer syntax. The two-handler workaround is not stylistic; it is the only form that survives the formatter.

## Why that matters past this PR

`hacs.json` declares `"homeassistant": "2024.1.0"`, and Home Assistant's Python floor only reached 3.14 very recently:

| Home Assistant | Requires-Python |
|---|---|
| 2025.1.0 | >=3.12.0 |
| 2025.6.0 | >=3.13.2 |
| 2026.1.0 | >=3.13.2 |
| 2026.7.4 | **>=3.14.2** |

So any `except (A, B):` written anywhere in this integration gets silently rewritten into a form that **fails to import** for every user on 2026.6 or older — while CI stays green, because CI runs 3.14. A config-flow module that cannot import breaks setup for all three protocols, not just the new one.

Nothing on `main` is currently affected — I grepped, there are no unparenthesised handlers today. This is a trap rather than a live bug, and it is one line to close:

```toml
[tool.ruff]
target-version = "py312"
```

Verified that this is sufficient:

```
no target-version:          except ValueError, TypeError:      -> 3.12 SyntaxError
--target-version py312:     except (ValueError, TypeError):    -> 3.12 parses OK
```

Worth its own PR rather than this one — it is a repo setting, and it would let the two handlers here collapse back into one if @trailro prefers that.

## On the PR itself

The validation-by-connecting argument is right, and it is right for a reason the original suggestion (mine, on #505) missed: **at the point the user picks a protocol there is no other transport to ask over.** Reading `http.enable` assumed a grill already connected, which is only true on reconfigure. Attempting the connection also covers the ESP-IDF line, where there is no `http` section to read at all — so one check answers for every grill instead of raising for some.

Treating "endpoint off" and "wrong address" as the same error is the right call too, and the message says so rather than pretending to know which.

`reconnect_on_poll` is the piece I would have got wrong. The other two transports reconnect themselves and a failed cycle is the correct response; HTTP's recovery *is* the next request, so refusing to talk to a disconnected transport strands it until a reload. Gating that on the protocol rather than doing it for everyone is right — and the comment naming why it would be unsafe for the websocket (a second receive loop, before pytboss#572 landed) is exactly the sort of thing that stops someone "simplifying" it later.

Two smaller things, neither blocking:

- `_start_api()` on the reconnect path calls `api.start()`, which re-resolves the grill spec through `get_grill` on every failed cycle — a thread hop and a catalogue lookup per poll while a grill is unreachable. Harmless, but `connect()` alone is what the path actually needs.
- `except (Unauthorized, RPCError)` is equivalent to `except RPCError`, since `Unauthorized` subclasses it. Naming both reads as though they take different paths. Worth either dropping `Unauthorized` or splitting them if the intent is to log differently.

## What I did not run

**The test suite does not run in my environment** — it fails identically on unmodified `main` (116 failed, 17 passed, `KeyError: 'pitboss'`), so I ran `main` as a control rather than reading anything into it. CI is the check here, and CI is green: build, mypy, ruff, Hassfest and HACS all pass.

I also have not exercised this against a grill. @trailro has, on a PB1600PS1 with `http.enable` true — which is the only way this feature gets real evidence, since @dknowles2's PBV4PS2 has the endpoint off and `pytest-socket` blocks the suite from reaching any grill in either direction.

## One interaction to sequence

This adds `"local": "Local (WiFi)"` alongside `"ble": "Bluetooth (local)"`, so two of the three options say "local". #369 renames the existing pair to plain **Bluetooth** and **Cloud**, which removes that collision — whichever lands second wants the third label reviewed against the other two rather than merged as-is. Both PRs also edit the same README section.

trailro:
Both small things answered — one in code, one with reasoning — plus a factual correction to the review that changes where the one-line repo fix goes.

**`except (Unauthorized, RPCError)` → `except RPCError`, at `b62252c`.** Right that the tuple read as two paths when it is one class; the handler's comment now carries what the tuple implied — the `rpc.auth_file` refusal is covered through its parent, and deliberately not routed to reauth, since it is HTTP-layer auth rather than the grill password. The failed-reconnect test keeps all three probe outcomes, the `Unauthorized` case landing through the subclass relationship the review pointed at.

**The `api.start()` on the reconnect path stays, with its cost named.** Agreed `connect()` alone is what the path needs — but pytboss's public surface has no connect-without-catalogue: `start()` is spec resolution plus `connect()`, and reaching for `api._conn.connect()` would put a private attribute in the coordinator. The actual overhead is one `Grill.from_dict` in an executor thread per failed cycle (the JSON read behind it is `@cache`d), at most once per backoff interval while a grill is unreachable. The right fix is a public `reconnect()` in pytboss — happy to send that upstream and swap it in here when a release carries it.

**Correction: this repo does have a ruff config.** `.ruff.toml` exists at the root with `target-version = "py314"` explicit on line 14 — added in #343, and its own header explains that `target-version` *used to* be inferred exactly as the review describes, which is why it was pinned. So the formatter behaviour is configured, not inferred, and the one-line fix is editing that line to `"py312"` rather than introducing a `[tool.ruff]` in a `pyproject.toml` this repo deliberately does not have (the header says that too). Conclusion unchanged, location matters: still worth its own PR, still the maintainer's call between lowering the target and raising the HACS floor.

**On the label sequencing:** agreed. Whichever of this and #369 lands second, I will reconcile the third label with the surviving pair — "Local" beside "Bluetooth"/"Cloud" if #369 wins the race, and #369's rename already removes the double-"local" collision this PR would otherwise create.

160 tests, ruff, `ruff format --check`, mypy clean on the updated branch.

dknowles2:
`b62252c` addresses the redundant handler. Confirmed `issubclass(Unauthorized, RPCError)` is `True`, so `except RPCError` covers both, and the new comment says which of the two it is covering and why neither is reauth's — that is the part worth having, since "answered badly" and "wrong password" landing in the same place looks like an oversight until someone explains it is not.

Re-verified at this head: `ruff check`, `mypy .` clean, `coordinator.py` parses under 3.12, and CI is green — build, mypy, ruff, Hassfest, HACS.

The `_start_api()` note I left is untouched, which is fine — I flagged it as non-blocking and it stays true either way: the reconnect path re-resolves the grill spec through `get_grill` on every failed cycle when `connect()` alone is what it needs. Worth a follow-up rather than another round here.

**Still blocked on @dknowles2's review, not on anything outstanding from me.** The syntax issue he raised is fixed and I have posted why CI was reporting honestly — his `CHANGES_REQUESTED` is his to lift.

dknowles2:
Re-checked at `af7b512` against the current `main` (`2ae517d`), which has moved a long way since my last look — #375 through #381 all landed in between.

**The README reconciliation is clean.** #375's safety bullet and its `pitboss.start_grill` entry both survived the rebase intact, alongside your three-protocol bullet and the Connection protocols section. Nothing was lost in either direction, which was the risk.

## The label collision is now concrete, and worth settling before either lands

```
this PR:  ble -> "Bluetooth (local)"   wss -> "WebSocket (cloud)"   local -> "Local (WiFi)"
#369:     ble -> "Bluetooth"           wss -> "Cloud"               (predates `local`)
```

Two of the three options in this PR say "local", which is the thing #369 set out to fix — "(local)" was doing the work "WebSocket" should have been doing, and now there is a genuinely local option to be confused with.

The end state that reads correctly is three plain words:

```json
"ble":   "Bluetooth",
"wss":   "Cloud",
"local": "Local"
```

Bluetooth is local too, so qualifying only one of them as "local" is what makes the pair ambiguous. "Local" alone is fine once "Cloud" is the thing it contrasts with — which is exactly what #369 establishes.

**Simplest path: take those three labels here**, and #369 becomes redundant and I close it. That avoids the sequencing question entirely, and it is one JSON object either way. If @dknowles2 would rather land #369 first, this PR just keeps the new `ble`/`wss` values and adds `local`.

Either is fine; the thing to avoid is shipping "Bluetooth (local)" next to "Local (WiFi)", since that is worse than what either PR started from.

## Still standing from my earlier review, still non-blocking

`_start_api()` on the reconnect path calls `api.start()`, which re-resolves the grill spec through `get_grill` on every failed cycle — a thread hop and a catalogue lookup per poll while a grill is unreachable. `connect()` alone is what the path needs. Worth a follow-up rather than another round here.

## Blocked only on @dknowles2

The `except` syntax he flagged is fixed and parses on 3.9/3.12; CI was reporting honestly (it pins `.python-version` = 3.14, where that syntax is legal). His `CHANGES_REQUESTED` is the only thing in front of this, and nothing is outstanding from me.

trailro:
Taken, at `0889639` — the three plain words:

```json
"ble":   "Bluetooth",
"wss":   "Cloud",
"local": "Local"
```

Your framing settles it: Bluetooth is local too, so qualifying only one of them was what made the pair ambiguous, and once "Cloud" is the contrast, "Local" alone reads correctly. Stored values untouched — labels only, nothing migrates. #369 is redundant with this in, as you said; close it when you take this.

Full suite green on the branch (173 passed, `python:3.14` container), JSON validated, ruff clean.

trailro:
Worth mentioning that pytboss had some fixes applied so would be good if both pytboss and ha-pitboss are bumped with this commit.

dknowles2:
`a788d06` fixes a real one, and it is a good catch on your own PR: re-validating on reconfigure meant a local entry could only ever be reconfigured while the grill was awake. Changing a password — the thing reauth sends people here to do — failed with `cannot_connect` for a reason that had nothing to do with the password.

Truth-tabled the new guard against every shape reconfigure can present:

```
local, same host, grill off      -> skip validation
local, host changed              -> validates
switching onto local             -> validates
switching off local              -> validates
local, host cleared              -> validates
local entry with no stored host  -> validates
wss untouched                    -> validates
```

All as intended. The `bool(host)` term is the one doing quiet work: without it, a local entry with an empty stored host submitted empty again would count as "unchanged" and skip straight past `host_required`. Worth keeping the comment near it, because it reads like a redundant truthiness check and is not.

`ruff check`, `mypy .` clean.

The reasoning is also the right one to have written down — "the stored address already proved itself when it was set, and validation exists to catch a *new* one". That is the distinction that makes skipping safe, rather than skipping because validation is inconvenient.

Still blocked only on @dknowles2's `CHANGES_REQUESTED`; nothing outstanding from either of us.

dknowles2:
`a2a7a01` is a merge of `main` and nothing else — no substantive change since `d469b8d`. Verified the merged result rather than assuming it, since `main` moved a fair way underneath this branch (#382–#387):

```
196 passed
ruff / ruff format --check / mypy — clean
```

Worth noting one collision that did **not** happen: #386 renamed `async_add_devices` to `async_add_entities` across the platform modules, and nothing in this branch reintroduced the old name.

The label collision I flagged earlier is also resolved. The three options now read **Bluetooth / Cloud / Local**, so only one of them says "local".

## Withdrawing one of my two earlier nits

I said `_start_api()` on the reconnect path re-resolves the grill spec through `get_grill` on every failed cycle, calling it "a thread hop and a catalogue lookup per poll". That was accurate as a description and misleading as a concern. Measured:

```
get_grill(): 0.61 ms per call
at a 10s poll on an unreachable grill: ~0.2 ms/hour of re-resolve
```

`_get_grills()` is `@cache`d, so the JSON load happens once and only `Grill.from_dict` repeats. Not worth changing, and not worth anyone's review time — my apologies for raising it as though it were.

The other nit is gone from the current head: `except (Unauthorized, RPCError)` is now a plain `except Unauthorized`.

## Still blocked on one thing

@dknowles2's `CHANGES_REQUESTED` from 2026-08-04T08:45. Not mine to clear, and the syntax error it was filed against is long fixed — the branch has been through several rounds since. Everything else here is green.

Separately, and not blocking this PR: the comment at `_async_poll` notes pytboss#572 is "not yet released". Still true — `main` there is 30 commits past `2026.8.4` and 2026.8.5 is an unpublished draft. That affects when the comment can be deleted, not whether this merges.

PR #372, opened 2026-08-04
Survive a null chamber reading instead of raising on it

trailro:
Found in a review pass over `main`; no issue filed, the defect and fix fit in one description.

The boards' own parsing routines put null on the wire for the no-reading sentinel (960) — `grillTemp` and `grillSetTemp` go through the same `convertTemperature` helper that gives an unplugged probe its `None`. The climate entity checked for the *key's presence* but coerced the value with `float()` unconditionally, so a chamber sensor that drops out (RTD fault, disconnect) made `current_temperature` raise `TypeError` on every state write until it came back. The probe sensors already pass the same value through as unknown; climate was alone in coercing.

Fix is the guard the probes effectively have: `None` reads as unknown, everything else as before. One new test drives both fields with null and asserts the attributes are simply absent — no test covered the sentinel value on the chamber fields before.

Verification: full suite green on Linux (134 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -101,14 +101,22 @@ class GrillClimate(BaseEntity, ClimateEntity):
 
     @property
     def current_temperature(self) -> float | None:
-        if (data := self.coordinator.data) and "grillTemp" in data:
-            return float(data["grillTemp"])
+        # Checked for None, not just presence: the boards' own parsing
+        # routines put null on the wire for the no-reading sentinel (960),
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Confirmed: `grillTemp`/`grillSetTemp` share the same `convertTemperature` no-reading sentinel (960 -> null) as the probe fields, which already get this guard — climate was the one place still coercing unconditionally with `float()`. Fix and test are correctly targeted.

PR #373, opened 2026-08-04
Back off on a pattern of failures, not a single miss

trailro:
A refinement of my own #366, offered as a decision rather than left as an oversight — the same courtesy #356 got.

#366 backs the poll interval off when a cycle fails, so a grill switched off behind the relay's ever-open socket stops eating a ten-second ping forever. Correct — but it triggers on the *first* failure, and one failure is as likely a mid-cook Bluetooth hiccup as a grill gone away. The cost lands exactly where freshness matters most: one lost ten-second poll during a cook became a sixty-second gap in temperatures.

This makes the trigger a pattern: two consecutive failures. The trade in numbers:

- **Mid-cook hiccup** (the common transient): before, a 60s data gap; now, the next poll comes 10s later as if nothing happened.
- **Grill actually off** (what #366 exists for): reaches the slow interval one failed poll later than before — 10 extra seconds of fast polling against a dead grill, once.

A success resets the count, so isolated failures never accumulate. `FAILURES_BEFORE_BACKOFF = 2` sits next to the two intervals in `const.py` with the reasoning; the existing backoff test now walks both cycles, and a new one pins that an isolated failure between successes never backs off.

Verification: full suite green on Linux (134 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/const.py
+++ b/custom_components/pitboss/const.py
@@ -20,6 +20,11 @@ MANUFACTURER = NAME
 # yet. They are the two values that would become fields when it grows one.
 ACTIVE_SCAN_INTERVAL = timedelta(seconds=10)
 STANDBY_SCAN_INTERVAL = timedelta(seconds=60)
+# Consecutive failed cycles before the poll interval backs off. One failure
+# is as likely a mid-cook hiccup as a grill gone away, and backing off on it
+# turns a lost ten-second poll into a sixty-second gap in the one situation
+# where freshness matters most. Two in a row is a pattern.
+FAILURES_BEFORE_BACKOFF = 2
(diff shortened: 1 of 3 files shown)


dknowles2 (approved):
Logic checked: `_failed_polls` resets to 0 only on the success path (after the try/except, never inside it), increments only in the `UpdateFailed` branch, and the backoff only fires once the count reaches `FAILURES_BEFORE_BACKOFF`. The two new/updated tests exercise exactly the two cases that matter (two-in-a-row backs off, a success in between resets). Reasoning in the comment matches the code.

PR #374, opened 2026-08-04
Hold the asked-for setpoint until the grill confirms it

trailro:
The climate card was the one control left out of the settle pattern. The board wipes its cached status the moment it forwards a command to the MCU, so the poll right after a set still carries the *old* setpoint — the unit select (#347) and the probe targets (#341) hold what was asked for `MCU_SETTLE_SECONDS` and then believe the grill again, but the climate slider snapped back to the stale value for up to a full poll cycle.

Same mechanism as its siblings, deliberately: a pending value shown until the grill reports it, one settle timer that a second set replaces rather than queues, a refresh when the window closes, and the grill's own answer winning after that — if the board never took the value, the entity stops asserting it.

One visible improvement beyond the hold: **the setpoint is snapped locally with the same arithmetic pytboss applies** (`min(accepted, key=...)` over the same list), so what is held and shown is the value the board will actually report rather than the one it silently corrects. The existing `test_set_temperature` now documents that — asking 275 of a board with a tens list sends and shows 270. Two new tests walk the hold: a stale poll inside the window does not snap the slider back, a confirming report hands control back to the grill, and an unconfirmed value expires with the window.

Stacked on #372 (`climate-survives-a-null-reading`) — both touch `climate.py`, and this builds on its guarded `target_temperature`. The diff includes that commit until it lands; I will rebase the moment it does, same as #356.

Verification: full suite green on Linux (136 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); `ruff check`, `ruff format --check`, `mypy .` clean.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -101,14 +101,22 @@ class GrillClimate(BaseEntity, ClimateEntity):
 
     @property
     def current_temperature(self) -> float | None:
-        if (data := self.coordinator.data) and "grillTemp" in data:
-            return float(data["grillTemp"])
+        # Checked for None, not just presence: the boards' own parsing
+        # routines put null on the wire for the no-reading sentinel (960),
+        # the same way an unplugged probe reads. `float(None)` would raise
+        # on every state write for as long as the chamber sensor is out.
+        if (data := self.coordinator.data) and (
+            temp := data.get("grillTemp")
+        ) is not None:
+            return float(temp)
         return None
 
     @property
     def target_temperature(self) -> float | None:
-        if (data := self.coordinator.data) and "grillSetTemp" in data:
-            return float(data["grillSetTemp"])
+        if (data := self.coordinator.data) and (
+            temp := data.get("grillSetTemp")
+        ) is not None:
+            return float(temp)
         return None
 
     async def async_set_temperature(self, **kwargs) -> None:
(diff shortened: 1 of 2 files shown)


dknowles2:
Reviewed and looks correct — verified the settle pattern matches the existing number/select entities exactly (`MCU_SETTLE_SECONDS`, single-timer replace-not-queue, `_handle_coordinator_update` clearing on confirmation) and the snap arithmetic matches `PitBoss.set_grill_temperature`'s own `min(accepted, key=...)` exactly, so what's held/shown is what the board will actually report.

Approving, but merging #372 first (to land the null-guard fix on its own) left this with a one-line docstring-placement conflict in `climate.py` — purely textual, no real disagreement in content. Rebase onto `main` should be a clean resolve. Happy to push that myself (maintainer-edits is on) if you'd rather not, or I can wait for you.

dknowles2 (approved):
Verified the settle pattern against the existing number/select entities (same MCU_SETTLE_SECONDS constant, single-timer replace-not-queue, clear-on-confirmation in _handle_coordinator_update) and the snap arithmetic against pytboss's own set_grill_temperature — both match exactly. Comment above has the merge-conflict note.

trailro:
Resolved — `main` merged in (append-only rather than a rewrite, as with #356) and the conflict was the one line you called: the docstring-and-pending block on `target_temperature`, kept on top of #372's guarded version. The diff against `main` is back to the settle change alone: `climate.py` +74, `test_climate.py` +109. Full suite green on the merged branch (137 passed, `python:3.14` container), ruff/format/mypy clean.

PR #375, opened 2026-08-04
Declare the floor and the flame the code actually has

trailro:
Two findings from a full three-pass audit of `main`, grouped because both are the same defect: what the repo *declares* diverged from what it *does*. No code changes.

**`hacs.json` admitted installs a year older than the code supports.** The floor said 2024.1.0, but the code needs `ConfigFlowResult` (2024.2), the coordinator `_async_setup` hook (2024.8 — on older HA it is silently never called, so `subscribe_state` never registers and push updates are dropped with no error), `_get_reauth_entry`/`_get_reconfigure_entry` and the implicit `OptionsFlow.config_entry` (2024.11), and `_abort_if_unique_id_mismatch` (~2024.12). A user on HA 2024.10 passes the HACS gate and then gets "Config flow could not be loaded" with nothing saying why. 2025.1.0 is the first round floor above all of them.

**The README's headline safety claim stopped being true when #370 merged.** "The integration cannot turn the grill on remotely" — it can, behind the opt-in, and the Actions section did not mention `pitboss.start_grill` at all. Anyone granting automations or a voice assistant access to the service surface on the strength of that sentence was working from a false premise. The bullet now states the real invariant — off by default, one deliberately-enabled action, the switch and climate card refuse either way — and the action is documented next to its sibling with the same preparation warning the service description carries.

Verification: full suite green on Linux (146 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); ruff, `ruff format --check`, mypy clean. Note #371 also edits the README; different sections, and whichever lands second reconciles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ dependency is updated.
 - **Reconfigurable, and it asks rather than gives up.** Change the model, password, or protocol without deleting the integration. If the grill starts rejecting the password, the integration asks you to re-enter it instead of retrying forever.
 - **Adapts to your grill.** The light, primer motor, recipe sensors, probe count, probe naming, and per-probe targets are all created from what your model and control board declare — not assumed.
 - **Polls faster when it matters.** Updates arrive quickly while the grill is running and back off in standby.
-- **Safety first.** The integration cannot turn the grill on remotely. You can monitor it and turn it off.
+- **Safety first, remote start off by default.** Out of the box the integration cannot light the grill: the power switch and climate card only ever turn it off. A deliberate opt-in in the integration options enables the `pitboss.start_grill` action -- and only that action; the switch and climate card refuse either way.
 
 ### Connection protocols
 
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both findings confirmed against `main`. Merging.

**The README's safety claim really is false now.** `main` still says "The integration cannot turn the grill on remotely" while `services.yaml` carries `start_grill:`, landed by #370. That is the worst kind of documentation error — a capability statement someone reasons about permissions from. Anyone reading that bullet and then handing a voice assistant blanket access to the `pitboss.*` service surface was working from a premise that stopped being true when #370 merged.

I checked the replacement text rather than taking it, since it makes a specific safety claim:

```
switch.py:72    "Not lighting the grill from here. Use the pitboss.start_grill ..."
climate.py:208  same, on HVACMode.HEAT
services.py:99  gated on entry.options.get(CONF_ENABLE_REMOTE_START, False)
```

So "the switch and climate card refuse either way", "opt-in", and "only that action" are each true as written.

**The HACS floor is demonstrably too low.** All five version-gated APIs are in use on `main`: `ConfigFlowResult` (10 sites), `_async_setup`, `_get_reauth_entry`, `_get_reconfigure_entry`, `_abort_if_unique_id_mismatch`. I confirmed `_async_setup` is a real `DataUpdateCoordinator` hook rather than something the repo defines itself, which is the one that worried me — a hook that simply never fires on an older HA is invisible, and it is where `subscribe_state` registers.

**What I did not verify:** the exact HA release each API was introduced in. I took 2024.2 / 2024.8 / 2024.11 / ~2024.12 from your description rather than checking release notes. It does not change the conclusion — 2025.1.0 clears all four either way, and 2024.1.0 clearly does not clear any of them — but the precise floor is your call rather than mine to certify.

Docs and JSON only; nothing to mutate. Noting #371 also edits the README in a different section, so whichever lands second reconciles.

PR #376, opened 2026-08-04
Let reconfigure keep the entries setup promises to keep

trailro:
From the same audit. Setup deliberately tolerates an entry from the board-remap era (#119/#315) whose stored model was never sold under its advertised prefix — `async_setup_entry` falls back to by-name resolution with a warning, "those keep working unchanged." Reconfigure did not mirror that tolerance, so exactly those entries hit a dead end in a first-class flow:

- **Unknown prefix** (`mygrill`-style IDs from the era): `_show_more_info_form` finds no models and hard-aborts the whole flow with `unknown_grill` — the owner cannot change a password or protocol at all.
- **Known board, stored model not on it**: the form's `vol.In` does not contain the entry's own model, so a working configuration cannot be re-saved, and every submittable choice moves the entry onto a different board's parsing routines.

The fix is one conditional: when a stored model exists and is not among the board's models, it is prepended to the choices. First-time setup is untouched — with no stored model the fallback cannot engage, and a genuinely unknown grill still aborts (the existing `test_user_flow_unknown_grill_aborts` pins that).

Two new tests, one per dead end: an unknown-prefix entry can re-save with a new password, and a stored model off its advertised board stays selectable and survives a round-trip.

Verification: full suite green on Linux (148 passed, `python:3.14` container); ruff, `ruff format --check`, mypy clean.

### DIFF
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -117,6 +117,17 @@ class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
         # since there is no entry yet whose setup could have warmed the
         # cache.
         models = await self.hass.async_add_executor_job(_models_on_board, control_board)
+        # Reconfigure mirrors the tolerance setup already has: an entry from
+        # the board-remap era can hold a model that was never sold under the
+        # advertised prefix, and `async_setup_entry` deliberately keeps those
+        # working by falling back to by-name resolution. Without this, that
+        # same entry hits a hard `unknown_grill` abort here (unknown prefix),
+        # or a model list its own stored model is not in -- so its owner
+        # could not re-save a working configuration, and the only
+        # submittable choices switched the entry onto a different board's
+        # parsing routines.
+        if isinstance(model, str) and model not in models:
+            models = [model, *models]
         if not models:
             return self.async_abort(
                 reason="unknown_grill",
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -229,6 +229,93 @@ async def test_reconfigure_flow_updates_entry(
 
 
 @pytest.mark.parametrize("model", ["PBV4PS2"])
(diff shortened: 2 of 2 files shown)


dknowles2 (approved):
Confirmed both dead ends and the asymmetry that causes them. Merging.

Setup really does promise this tolerance, in as many words — `__init__.py:127-134`:

```python
except InvalidGrill:
    # An entry from the era of board remapping can hold a model that was
    # never sold under the advertised prefix. Resolving by name alone is
    # what every release so far did, so those keep working unchanged.
```

And reconfigure went through `_models_on_board` with no such fallback, so the same entry that setup deliberately keeps alive could not be re-saved. That is the sharp edge: not a feature that fails, but a feature that works everywhere except the one flow you reach for when you need to change something.

The second dead end is the worse of the two. An abort at least tells the owner something is wrong; a form whose dropdown silently excludes their own model presents every remaining choice as valid, and each one moves the entry onto a different board's parsing routines. Picking one looks like configuring and is actually corrupting.

## The guard I checked

`isinstance(model, str)` is what keeps first-time setup's abort working, since `_show_more_info_form`'s default is `vol.UNDEFINED`:

```
vol.UNDEFINED is a str? False
```

So on a fresh flow the prepend cannot engage, `models` stays empty for a genuinely unknown prefix, and the existing `test_user_flow_unknown_grill_aborts` still means what it says. Worth confirming rather than assuming — an `if model and ...` would have looked equivalent and quietly broken that.

Placing it before the `if not models` abort rather than after is what makes the unknown-prefix case work at all; after it, the abort would still fire first.

`ruff check`, `mypy .` clean; CI green — build, mypy, ruff, Hassfest, HACS.

## What I did not run

The suite does not run in my environment (it fails identically on unmodified `main`), so I have not executed your two new tests. CI has. My verification is the setup/reconfigure asymmetry, the `vol.UNDEFINED` guard, and the static checks.

PR #377, opened 2026-08-04
Reload on options changes only when an option asks for it

trailro:
From the same audit — a refinement of my own listener from #370, offered as a decision the way #373 refined #366.

The options listener reloaded the entry on every save. A reload is not free: **on Bluetooth it waits for the grill to advertise again**, up to the 30s `_connect_ble` window and then `ConfigEntryNotReady` retries — so flipping the remote-start toggle while the grill was asleep, which is exactly the prepare-for-later case that toggle exists for, took the whole integration down, entities and all, until the grill next woke up. And the one option that exists is read live at its point of use, so the reload bought nothing.

The listener stays registered, and the trap it guarded against stays guarded — just declaratively: `RELOAD_OPTIONS` names the options that fix something at entity registration, the listener reloads when the set is non-empty, and its docstring says both why it is empty today and what belongs in it tomorrow. A future option has one obvious place to declare itself, next to the reasoning about what an unconditional reload costs.

New test pins the behaviour: saving a live-read option on a loaded entry leaves the entry loaded and the coordinator object identity unchanged. The service-gating tests already prove the flag is honoured without a reload.

Verification: full suite green on Linux (147 passed, `python:3.14` container); ruff, `ruff format --check`, mypy clean.

### DIFF
--- a/custom_components/pitboss/__init__.py
+++ b/custom_components/pitboss/__init__.py
@@ -172,15 +172,27 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     return True
 
 
-async def _async_options_updated(hass: HomeAssistant, entry: ConfigEntry) -> None:
-    """Reload when the options change.
+RELOAD_OPTIONS: frozenset[str] = frozenset()
+"""Options whose change requires reloading the entry.
 
-    The remote-start toggle is read live, so this is not strictly needed for
-    it -- but an options flow without a listener is a trap for the next
-    option added, and Home Assistant fixes some entity properties (a display
-    unit, for one) when the entity is registered.
-    """
-    await hass.config_entries.async_reload(entry.entry_id)
+Empty because every option that exists today -- the remote-start toggle --
+is read live where it is used. An option that fixes something at entity
+registration (a display unit, a polling interval baked into the
+coordinator) belongs in this set when it is added.
+
+The set exists because reloading unconditionally has a real cost: on
+Bluetooth a reload waits for the grill to advertise again, so flipping a
+toggle while the grill was asleep took the whole integration down --
+entities and all -- until the grill next woke up. The listener stays
+registered so a future option cannot forget to think about this; it just
+reloads only when the option asks for it.
+"""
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both premises verified. Merging.

**The option really is read live** — `services.py:99` reads `entry.options.get(CONF_ENABLE_REMOTE_START, False)` at call time, so the reload changed nothing about whether the toggle is honoured.

**The reload cost is exactly as described** — `__init__.py`:

```python
try:
    async with timeout(30):
        async with ready:
            await ready.wait_for(conn.is_connected)
except TimeoutError:
    raise ConfigEntryNotReady
```

So on Bluetooth a reload blocks on the grill advertising, for up to 30 seconds, and then hands the entry to `ConfigEntryNotReady` retries. Entities gone until the grill wakes.

The irony is worth stating because it is what makes this more than a tidy-up: the remote-start toggle exists for the prepare-for-later case — you enable it *before* you intend to light the grill — which is precisely when the grill is most likely to be asleep. The one option in the integration was the one whose most natural use took the integration down.

Keeping the listener registered rather than deleting it is the right call, and it is the same judgment as #373 refining #366: the guard was correct in intent and wrong in cost, so narrow it rather than remove it.

`ruff check`, `mypy .` clean; CI green.

## One note on the granularity, not blocking

`if RELOAD_OPTIONS:` reloads whenever the **set is non-empty**, not when a changed option is in it. So the day someone adds a reload-requiring option, flipping the remote-start toggle starts reloading again and the 30-second Bluetooth cost comes back for it — the exact case this PR removes.

The docstring is honest about this ("reloads when the set is non-empty"), and I can see why: Home Assistant's options listener does not hand you which key changed without tracking previous values yourself, so set-non-emptiness is the cheap proxy. Worth knowing it is a proxy, since the title reads as per-option.

If it ever matters, comparing against a stashed copy of `entry.options` in the listener is the usual way, and the place to put that is this function.

## What I did not run

The suite does not run in my environment (fails identically on unmodified `main`), so I have not executed the new test. CI has.

PR #378, opened 2026-08-04
Agree with the command path on what an absent unit flag means

trailro:
From the same audit. Four read sites defaulted a missing `isFahrenheit` to **Celsius** — `coordinator.grill_unit`, `climate.temperature_unit`, the probe sensor's unit, and the unit select — while pytboss's `_is_fahrenheit`, the default every setpoint is snapped with, reads it as **Fahrenheit**.

The window where this matters is real, not theoretical: on live parsing code the flag is emitted by the temperatures frame on all 20 boards but by the status frame on only 3 (the other 17 carry it commented out — a raw grep says otherwise, which is exactly how an earlier pass of mine wrongly cleared this). The board wipes each frame as it forwards commands, so a poll can land a status frame alone: a non-empty state without the key. For that window the entities presented Celsius units, Celsius bounds, and Celsius min/max — and a setpoint sent through them was snapped by pytboss against the Fahrenheit list. Ask for 100 °C, the board is sent 180 °F.

All four sites now default to Fahrenheit, matching the library, with the reasoning kept at the one place the divergence lived (`coordinator.grill_unit`'s docstring). Tests cover the window on each surface: climate's step reads 5.0, a raw probe 165 stays 165 °F instead of being converted as 329, and the select reports °F.

Verification: full suite green on Linux (148 passed, `python:3.14` container); ruff, `ruff format --check`, mypy clean.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -118,7 +118,10 @@ class GrillClimate(BaseEntity, ClimateEntity):
 
     @property
     def temperature_unit(self) -> str:
-        if (data := self.coordinator.data) and not data.get("isFahrenheit", False):
+        # Absent defaults to Fahrenheit -- the same default pytboss snaps
+        # setpoints with, so the unit shown and the unit commanded agree
+        # even on a state built from a status frame alone.
+        if (data := self.coordinator.data) and not data.get("isFahrenheit", True):
             return UnitOfTemperature.CELSIUS
         return UnitOfTemperature.FAHRENHEIT
 
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -94,8 +94,17 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
 
     @property
     def grill_unit(self) -> str:
-        """The unit the grill is currently working in."""
-        if (data := self.data) and not data.get("isFahrenheit"):
+        """The unit the grill is currently working in.
+
+        Absent defaults to Fahrenheit, matching pytboss's `_is_fahrenheit` --
+        the default the command path snaps setpoints with. The key is only
+        emitted by the temperatures frame on most boards, so a state built
+        from a status frame alone (a poll landing right after the firmware
+        wipes the other half) is non-empty without it; reading that as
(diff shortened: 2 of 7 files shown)


dknowles2 (approved):
Confirmed, and the window is wider than the description claims. Merging.

I re-derived the emission counts with pytboss's own `_live_code` rather than a grep, precisely because you flagged that trap:

```
reachable boards: 17
  temperatures frame emits isFahrenheit (live): 17/17
  status frame emits isFahrenheit (LIVE):        2/17  -> LBL, LFS
  status frame per a RAW GREP:                  17/17
  => the grep overstates by 15 boards
```

Your "3 of 20" is the declared count; on the **17 reachable** boards it is **2** — LBL and LFS. So **15 of 17** boards can produce a non-empty state with no `isFahrenheit` at all, which is the window this bug lives in. Bigger than stated, and the grep genuinely does say the opposite, exactly as you warned.

The divergence is real on both sides:

```
pytboss:     self._state.get("isFahrenheit", True) is not False   -> absent = Fahrenheit
ha-pitboss:  not data.get("isFahrenheit")                         -> absent = Celsius
```

Note `climate.temperature_unit` was the loudest of the four — it passed an explicit `False` default, so it was not an oversight of omission but a stated choice in the wrong direction.

`ruff check`, `ruff format --check`, `mypy .` clean; CI green — build, mypy, ruff, Hassfest, HACS.

## Why this one is worth the fuss

The failure is silent and wrong rather than absent. During that window the card offers Celsius bounds and a 5.0 step, the user asks for 100, and pytboss snaps it against the Fahrenheit list — the grill gets 180 °F. Nothing errors, nothing logs, and the reading that comes back a second later is in the other unit, so it does not even look inconsistent for long.

Defaulting to the library's answer rather than picking one independently is the right resolution. The alternative — making pytboss default to Celsius — would be worse: `_is_fahrenheit` also feeds `set_grill_temperature`'s unit flag, which every board's command routine reads, and that is the surface with hardware behind it.

## What I did not run

**The test suite does not run in my environment** (it fails identically on unmodified `main`), so I have not executed the three new tests. CI has, and it is green. My verification here is the emission sweep, the two defaults, and the static checks.

Nor have I seen this on a grill — the window needs a poll to land between the two frames, which is the same interleaving #326 fixed on the merge side.

PR #379, opened 2026-08-04
Make the services say what actually happened

trailro:
Two error surfaces from the audit's minor queue, one file, one theme: the service layer answering with the wrong words.

**`start_grill` sent the ignite command unguarded.** It is authenticated like every other write, so a stale password surfaced as "Unknown error" with a traceback and no path to fix it — while its sibling `set_grill_password` already maps `Unauthorized` to the reauth flow plus a clean message. It now gets the identical treatment, and any other failure is wrapped as a service error naming the operation. Deliberately mirrored line for line rather than shared, so the two stay greppable next to their own gates.

**`_coordinator_for_device` called a real grill "not a PitBoss grill".** An entry mid-`ConfigEntryNotReady` (grill off) or disabled has no coordinator in `hass.data`, but its device and entry are right there — the right target in the wrong state, which is not the caller's fault. The resolver now distinguishes: a device with a PitBoss entry but no live coordinator gets "the grill is not loaded yet" as a `HomeAssistantError`; the validation error is reserved for devices that genuinely are not ours.

Three new tests: `start_grill` rejection opens reauth, other failures wrap with the operation named, and an unloaded-but-real grill is named as such.

Verification: full suite green on Linux (151 passed, `python:3.14` container — macOS cannot run the suite natively per AGENTS.md); ruff, `ruff format --check`, mypy clean.

### DIFF
--- a/custom_components/pitboss/services.py
+++ b/custom_components/pitboss/services.py
@@ -50,11 +50,23 @@ def _coordinator_for_device(
     device = dr.async_get(hass).async_get(device_id)
     if device is None:
         raise ServiceValidationError(f"Unknown device: {device_id}")
+    ours = False
     for entry_id in device.config_entries:
-        coordinator = hass.data.get(DOMAIN, {}).get(entry_id)
         entry = hass.config_entries.async_get_entry(entry_id)
-        if coordinator is not None and entry is not None:
+        if entry is None or entry.domain != DOMAIN:
+            continue
+        ours = True
+        coordinator = hass.data.get(DOMAIN, {}).get(entry_id)
+        if coordinator is not None:
             return coordinator, entry
+    if ours:
+        # The right target in the wrong state, which is not the caller's
+        # fault -- an entry mid-retry because the grill is off, or disabled.
+        # "Not a PitBoss grill" here sent people debugging their automation
+        # instead of their grill.
+        raise HomeAssistantError(
+            "The grill is not loaded yet; it may be off or still connecting."
+        )
     raise ServiceValidationError(f"Device {device_id} is not a PitBoss grill")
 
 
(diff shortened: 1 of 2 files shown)


dknowles2 (approved):
Both surfaces verified against the sibling they claim to mirror. Merging.

`set_grill_password` does exactly what you describe — `Unauthorized` → `entry.async_start_reauth(hass)` plus the same sentence, everything else wrapped naming the operation. `start_grill` now matches line for line. Mirroring rather than extracting a helper is the right call here: both live directly under their own gates, and a shared wrapper would put distance between the gate and what it does on failure, in the one service where that gate is a safety property.

`ruff check`, `ruff format --check`, `mypy .` clean.

## The resolver change is better than the description says

The stated fix is distinguishing "not loaded" from "not ours". It also fixes how "ours" is decided at all. Before:

```python
coordinator = hass.data.get(DOMAIN, {}).get(entry_id)
entry = hass.config_entries.async_get_entry(entry_id)
if coordinator is not None and entry is not None:
```

Ownership was inferred from a `hass.data` lookup succeeding. That happened to be safe — a foreign `entry_id` is not a key under `DOMAIN` — but it conflated "this entry is ours" with "this entry is loaded", which is precisely the conflation causing the bad message. The new `entry.domain != DOMAIN` asks the question directly, so the two facts are separable and the error can be accurate.

The exception choice is right and worth naming, since the two are easy to swap: `ServiceValidationError` is Home Assistant's "the caller got this wrong", `HomeAssistantError` is "the call was fine, the world was not". A grill that is off is the second, and it was answering with the first — which is why it sent people to debug their automation. (`ServiceValidationError` subclasses `HomeAssistantError`, so anything catching the broader type is unaffected either way.)

## What I did not run

The suite does not run in my environment (fails identically on unmodified `main`), so I have not executed your three new tests. CI has, and my checks here are the sibling comparison, the imports, the exception hierarchy and the static tools.

PR #380, opened 2026-08-04
Find the grill, and let it be added

trailro:
Two discovery-and-identity gaps from the audit, one theme: a supported grill failing to get in the door.

**Manual device-ID input stored whatever was typed.** Real IDs are the uppercase advertised name, and the board lookup behind the next step compares names case-sensitively — a pasted `" pbl-3b22cd "` hard-aborted the flow with `unknown_grill` for a fully supported grill, discarding the input. Manual entry is now stripped and uppercased; the Bluetooth path already arrives normalized, and the entry-data assertion tests pin the canonical form.

**The Bluetooth matcher list had drifted from the catalogue in both directions.** `PBX1-*` matched a board whose only model is in `UNSUPPORTED_MODELS`, so a discovered PBX1 always walked bluetooth → confirm → dead end. And four supported boards — PBC2, PBL3, PBM2, PBV2 — had no matcher at all (`PBM-*` does not glob `PBM2-…`), so those grills were never discovered and had to be typed in by hand. The list now matches the catalogue exactly.

The part with lasting value is the new `test_manifest.py`: it derives the expected matcher set from the pytboss catalogue and compares it to the manifest, so the next weekly definitions refresh that adds or retires a board fails the build instead of drifting silently. It caught its first mistake before this PR was ever pushed — PBVA looked eligible by raw parsing routine, but both its models are in `UNSUPPORTED_MODELS`, so it gets no matcher; I had it in the list until the test said no.

Verification: full suite green on Linux (150 passed, `python:3.14` container); ruff, `ruff format --check`, mypy clean. Note this edits `manifest.json`'s `bluetooth` array only — the `requirements` pin is untouched, per AGENTS.md.

### DIFF
--- a/custom_components/pitboss/config_flow.py
+++ b/custom_components/pitboss/config_flow.py
@@ -81,7 +81,11 @@ class PitBossFlowHandler(ConfigFlow, domain=DOMAIN):
                 data_schema=vol.Schema({vol.Required(CONF_DEVICE_ID): str}),
             )
 
-        self._device_id = user_input[CONF_DEVICE_ID]
+        # Typed by hand, so normalized: real device IDs are the uppercase
+        # advertised name (prefix + MAC-derived hex), and the board lookup
+        # behind the next step compares names case-sensitively -- a pasted
+        # " pbl-3b22cd " aborted the flow for a fully supported grill.
+        self._device_id = user_input[CONF_DEVICE_ID].strip().upper()
         await self.async_set_unique_id(self._device_id.lower())
         self._abort_if_unique_id_configured()
         return await self.async_step_more_info()
--- a/custom_components/pitboss/manifest.json
+++ b/custom_components/pitboss/manifest.json
@@ -24,6 +24,10 @@
     },
     {
       "connectable": false,
+      "local_name": "PBC2-*"
+    },
+    {
+      "connectable": false,
       "local_name": "PBD-*"
     },
     {
@@ -44,7 +48,15 @@
     },
     {
       "connectable": false,
+      "local_name": "PBL3-*"
+    },
+    {
+      "connectable": false,
       "local_name": "PBM-*"
+    },
+    {
+      "connectable": false,
+      "local_name": "PBM2-*"
     },
     {
       "connectable": false,
@@ -60,7 +72,7 @@
     },
     {
       "connectable": false,
-      "local_name": "PBX1-*"
+      "local_name": "PBV2-*"
     }
   ],
   "codeowners": [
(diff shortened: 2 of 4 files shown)


dknowles2 (approved):
Derived the expected matcher set from the catalogue myself rather than reading your test, and it agrees exactly:

```
MISSING from manifest: ['PBC2-*', 'PBL3-*', 'PBM2-*', 'PBV2-*']
EXTRA in manifest:     ['PBX1-*']
PBX1: models=['PBX - test 1']          unsupported=['PBX - test 1']
PBVA: models=['PBV30DS','PBV30DX']     unsupported=both
PBL2: models=['PB1150DX','PB850DX']    unsupported=none
after this PR: matchers=18  expected=18  exact match=True
```

Four missing, one spurious, and PBVA correctly absent. The `requirements` pin is untouched.

## My first derivation was wrong in the exact way your test exists to prevent

I built the expected set from `grills.get_grills()` boards — and it told me `PBL2-*` was also spurious. It is not: `get_grills()` yields one variant per model, so a board reachable only via an explicit `control_board` argument never appears in it. PBL2 carries PB1150DX and PB850DX, both supported, both listed under PBL3 as the vendor's later board.

Had I trusted that, I would have told you to delete a matcher that four real grills depend on for discovery. The catalogue has to be read from the raw entries, not from the default-resolution helper — which is precisely the distinction a hand-maintained list gets wrong, and precisely why deriving it in a test is worth more than this PR's own fix.

Your note that the test caught PBVA before you pushed is the same story from the other side. That is two independent people getting the eligibility rule wrong in one PR, and the test catching both.

## The two gaps

`PBX1-*` is the visible one — a discovered PBX1 walked bluetooth → confirm → dead end, because its only model is `PBX - test 1`, a vendor test entry.

The four missing matchers are the quieter and worse one. `PBM-*` does not glob `PBM2-…`, so a PBC2, PBL3, PBM2 or PBV2 owner never saw their grill discovered at all — it simply looked unsupported, with nothing to tell them manual entry would work. Between that and the manual-entry normalisation, this PR fixes both halves of the same experience: the grill is not found, and typing it in by hand does not work either.

`.strip().upper()` is right, and the case-sensitivity is real — the lookup behind the next step compares names exactly, and the Bluetooth path already arrives normalized, so this only touches the hand-typed path.

`ruff check`, `mypy .` clean; CI green.

## What I did not run

The suite does not run in my environment (fails identically on unmodified `main`). CI has run your new `test_manifest.py`; my check above is an independent derivation of the same property.

PR #381, opened 2026-08-04
Invalidate values a unit change leaves meaningless

trailro:
Three related fixes from the audit's minor queue, one theme: numbers held past the frame that gave them meaning. Plus one statistics one-liner riding along, #359-style.

**The settle windows held their value through a grill unit flip.** Climate and the probe targets show the asked-for value until the grill confirms it — but a flip of the unit select (or the panel) inside the window left 225 from a Fahrenheit set presented as **225 °C**, above the entity's own bounds, until the window expired. The pending value now records the unit it was captured in and is dropped the moment the grill reports a different one — the grill's own report wins immediately instead of six seconds later.

**`restored_targets` stored bare numbers with no unit.** Captured in whatever unit the grill had at the time, seeded in whatever unit it had at ignition: 74 saved while the grill spoke Celsius became 74 °F after a panel flip. The held values are now unit-fixed to Fahrenheit — the same convention as the grill's own virtual-data store — converted at capture and again at seeding, so 74 °C seeds as 165 °F whatever happened to the panel in between.

**The rider:** `recipeStep` loses its `MEASUREMENT` state class for exactly the reason `recipeTime` did in #359 — long-term statistics of a step index mean nothing. Same orphaned-statistics repair note applies.

Tests cover each: a unit flip inside the window snaps climate and the probe target to the grill's report rather than the stale number; a target captured under Celsius reads and seeds as 165 after a flip to Fahrenheit (asserted on the actual `set_probe_target` call); and `recipe_step`'s state carries no `state_class`.

Verification: full suite green on Linux (151 passed, `python:3.14` container); ruff, `ruff format --check`, mypy clean.

### DIFF
--- a/custom_components/pitboss/climate.py
+++ b/custom_components/pitboss/climate.py
@@ -58,6 +58,7 @@ class GrillClimate(BaseEntity, ClimateEntity):
         )
         self._attr_unique_id = f"{self.entity_description.key}_{entry_unique_id}"
         self._pending_target: float | None = None
+        self._pending_unit: str | None = None
         self._cancel_settle: CALLBACK_TYPE | None = None
 
     async def async_added_to_hass(self) -> None:
@@ -158,9 +159,15 @@ class GrillClimate(BaseEntity, ClimateEntity):
     @callback
     def _handle_coordinator_update(self) -> None:
         if self._pending_target is not None:
-            reported = (self.coordinator.data or {}).get("grillSetTemp")
-            if reported is not None and float(reported) == self._pending_target:
+            # A unit change invalidates the number outright: 225 held from a
+            # Fahrenheit set must not be presented as 225 C for the rest of
+            # the settle window. The grill's own report wins immediately.
+            if self.temperature_unit != self._pending_unit:
                 self._pending_target = None
+            else:
+                reported = (self.coordinator.data or {}).get("grillSetTemp")
+                if reported is not None and float(reported) == self._pending_target:
+                    self._pending_target = None
         super()._handle_coordinator_update()
 
     async def async_set_temperature(self, **kwargs) -> None:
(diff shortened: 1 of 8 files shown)


dknowles2 (approved):
All three verified. Merging. `ruff check`, `mypy .` clean; CI green.

The settle-window fix is the right shape — recording the unit alongside the value, rather than trying to convert the pending number, is correct because the number was never wrong, it just stopped meaning anything. Invalidating beats converting here.

**The restore chain is complete**, which is the part I went looking for a hole in. `async_added_to_hass` already reads `last_data.native_unit_of_measurement` and converts to the grill's current unit before handing off, and `note_restored_target` then converts that to Fahrenheit. So the path is `stored (with its unit) → grill unit → Fahrenheit`, with no double conversion, and a panel flip while Home Assistant was down is handled. I had assumed that path would be the gap and it is not.

**Round-tripping is stable**, which matters because these values survive repeated flips:

```
unstable Celsius values in 30..120: 0
74 C -> [74, 74, 74, 74, 74]   (four flips)
```

No drift, so a held target does not walk a degree per flip.

## One seam worth knowing about, not blocking

`_from_fahrenheit` uses Home Assistant's `round`; the boards use `ftoc`'s `floor`. On PBA, `p1Target` **is** run through `ftoc`, so the same underlying value is reported differently by the two:

```
held 165 F -> HA displays  round((165-32)/1.8) = 74 C
              grill reports floor((165-32)/1.8) = 73 C
```

89 of the 200 Fahrenheit values in 100..300 diverge. Two consequences, both small:

- while the value is held, the card shows 74; once the grill reports the same target, it becomes 73 — a one-degree jump that looks like the grill disagreeing.
- the number entity clears its settle window on `probe_target(...) == self._pending_value`, which cannot match across that seam, so the window expires on its timer instead of confirming early.

This is the same family as #327 and as the `floor`/`round` gap I raised on pytboss#593 — the boards truncate where the obvious implementation rounds, and every place that mirrors their arithmetic has to truncate too. Using `math.floor((f - 32) / 1.8)` in `_from_fahrenheit` would close it, at the cost of disagreeing with `TemperatureConverter` — which is arguably the right trade here, since the thing being matched is the board, not Home Assistant.

Not blocking, and genuinely a judgement call rather than a defect.

## The rider

`recipeStep` losing `MEASUREMENT` is right for the reason `recipeTime` did — long-term statistics of a step index are meaningless. Same orphaned-statistics note applies as #359.

## What I did not run

The suite does not run in my environment (fails identically on unmodified `main`). CI has run your tests; my checks here are the restore chain, the round-trip stability, and the `ftoc` comparison above.

PR #384, opened 2026-08-05
Cover the probe-target restore and settle window

dknowles2:
Stacked on #383, which is what makes the suite runnable locally in the first place. Merge that first; this diff is the two test blocks only.

## Two behaviours had no test

Deleting either of these left the suite green:

| Code | What it does |
|---|---|
| `number.py` restore block | Restores a probe target across a restart, converting if the grill's unit changed while Home Assistant was down |
| `number.py` `_async_confirm` | The MCU settle window that stops the slider holding a target the grill never took |

## The restore case is the sharper one

`restored_targets` is kept in **Fahrenheit**; the entity converts the stored value into the grill's *current* unit before handing it over, and the coordinator converts from there. Both halves have to be right.

A target stored as `71 °C` and restored onto a grill now reporting Fahrenheit has to arrive as **160**. Without the conversion it is filed as `71 °F` — below the bottom of the probe range, and no warmer than the meat starts out. Nothing would have caught that.

## Mutation results

Every one of these survived before and fails now:

```
drop the restore unit conversion                    caught
restore: round -> truncate                          caught
restore: convert the wrong way                      caught
restore: ignore the grill's own target              caught
settle: never refresh                               caught
settle: keep holding the value                      caught
settle: never arm the timer                         caught
settle: queue a second timer instead of replacing   caught
```

`number.py` goes from **84% to 99%** branch coverage; the module total from 94% to 95%.

## One thing worth knowing about the settle tests

The first version of the "second set replaces the window" test passed but **did not discriminate** — the mutation that queues a second timer instead of cancelling the first still survived it. The timers are armed against loop time, so passing a future datetime to `async_fire_time_changed` does not actually move the two windows apart: both expire on the same fire, and a queued second timer looks identical to a replaced one.

It uses `freezer` now, and catches the mutation. Worth remembering for any future timer test in this repo.

## Not in this PR

`climate.py`'s `min_temp`/`max_temp` are still uncovered, and are **not** a test gap — see the issue I am filing alongside this. Those lines are unreachable for every grill in the catalogue, so a test would have to build a spec that does not exist.

### DIFF
--- a/tests/test_number.py
+++ b/tests/test_number.py
@@ -1,13 +1,21 @@
 from collections.abc import Awaitable, Callable
+from datetime import timedelta
 from unittest.mock import Mock
 
 import pytest
 from conftest import get_entity
-from homeassistant.core import HomeAssistant
+from freezegun.api import FrozenDateTimeFactory
+from homeassistant.const import UnitOfTemperature
+from homeassistant.core import HomeAssistant, State
+from homeassistant.util import dt as dt_util
 from pytboss.exceptions import UnsupportedOperation
-from pytest_homeassistant_custom_component.common import MockConfigEntry
+from pytest_homeassistant_custom_component.common import (
+    MockConfigEntry,
+    async_fire_time_changed,
+    mock_restore_cache_with_extra_data,
+)
 
-from custom_components.pitboss.const import DOMAIN
+from custom_components.pitboss.const import DOMAIN, MCU_SETTLE_SECONDS
 from custom_components.pitboss.coordinator import PitBossDataUpdateCoordinator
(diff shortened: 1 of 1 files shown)


dknowles2:
**No CI here, and that is expected, not broken.** `build-and-test.yml` and `validate.yml` are both scoped to `branches: ["main"]`, so a PR targeting `fix-macos-test-suite` runs nothing. GitHub will retarget this to `main` automatically when #383 merges, and CI runs then.

Verified locally in the meantime, on the merged result: 165 passed, `ruff check`, `ruff format --check` and `mypy .` all clean.

PR #387, opened 2026-08-05
Let the bounds be the setpoint list, with nothing behind them

### LINKED ISSUE #385: climate min_temp/max_temp fallback is unreachable for every catalogued grill
Found while filling coverage gaps in #384. This one is not a missing test — the lines cannot be reached.

## The code

`climate.py`:

```python
@property
def min_temp(self) -> float:
    if accepted := self._accepted_setpoints:
        return min(accepted)
    from_unit = UnitOfTemperature.FAHRENHEIT
    to_unit = self.temperature_unit
    if (min_temp := self.coordinator.api.spec.min_temp) is None:
        min_temp = DEFAULT_MIN_TEMP
    return TemperatureConverter.convert(min_temp, from_unit, to_unit)
```

`max_temp` is the same shape. Everything after the walrus is uncovered, in both.

## Why it never runs

`_accepted_setpoints` is `coordinator.accepted_setpoints(unit)`, which is empty only when `spec.temp_increments` is empty. Checked against every entry in the catalogue the manifest pins (pytboss 2026.8.4):

```
entries: 147
accepted_setpoints EMPTY in F: 0
accepted_setpoints EMPTY in C: 0
```

Celsius cannot be empty either — `accepted_setpoints` derives it from the Fahrenheit list when `celsius_temp_increment` is null, so a grill with Fahrenheit increments always has Celsius ones.

So `min_temp`/`max_temp` return from the first line for every grill this integration supports, in both units, always.

## And the inner fallback is dead twice over

`DEFAULT_MIN_TEMP` / `DEFAULT_MAX_TEMP` guard against `spec.min_temp` being `None`, which does happen — three models declare `min_temp: "Smoke"` and `max_temp: "High"`, which pytboss's `int()` cannot parse:

```
spec.min_temp is None for: PB1000PL, PB1285KC, PB2180LK
```

The reason for the guard is real. It just sits behind a branch that is never taken.

## What to do with it

Not obvious, which is why this is an issue rather than a PR:

1. **Delete the fallback**, and let `min_temp`/`max_temp` be `min(accepted)`/`max(accepted)`. Smallest code, but it assumes the catalogue never ships a grill without increments, and it would `ValueError` on an empty sequence rather than degrade if one ever did.
2. **Keep it and test it** against a synthetic spec. Honest about the intent, but the test asserts on a grill that does not exist, which is the kind of test that outlives its reason.
3. **Leave it and note it.** Cheapest, and the note is the point — right now the comment history gives no hint that these lines are unreachable, so the next person to read them will assume they matter.

My preference is 1, on the grounds that a grill with no setpoint list would break far more than the slider bounds — `accepted_setpoints` feeds `extra_state_attributes` and the setpoint snapping too, so degrading gracefully *here alone* buys nothing.

Worth checking against a stance rather than my guess, since it touches what a user's climate card offers.

trailro:
Fixes #385, taking option 1 — with the argument from the issue thread that upgrades it from a coverage call to a structural one: a grill without increments cannot be constructed (`Grill.from_dict` raises on the empty increment string at catalogue load), so the graceful degradation this path promised was impossible, not merely untested. A vendor-data regression would be loud at setup either way, and — as the issue itself notes — `accepted_setpoints` feeds `allowed_setpoints` and the snap too, so degrading bounds alone bought nothing.

`min_temp`/`max_temp` are now `min`/`max` of the accepted list, full stop, with the reasoning recorded where the fallback used to live so the next reader knows why nothing is behind them. The three Smoke/High models (PB1000PL, PB1285KC, PB2180LK) stay covered by the existing parametrized bounds tests — PB2180LK's 160/500 asserts the only path there is now. `DEFAULT_MIN_TEMP`/`DEFAULT_MAX_TEMP` leave `const.py` and the `TemperatureConverter` import leaves `climate.py`, both consumerless; the stale test comment that still described the fallback is reworded.

Verification: full suite green in the `python:3.14` container (180 passed) **and, for the first time, natively on macOS** thanks to #384's `mock_bluetooth_adapter_history` fixture — 180 passed with one teardown-only error in the usb/inotify polling that predates this change. ruff, `ruff format --check`, mypy clean.


dknowles2:
Verified independently rather than taking the description on trust, since the whole change rests on one claim: that a grill without increments cannot exist.

**It holds.** `Grill.from_dict` rejects every degenerate form, against the catalogue the manifest pins:

```
catalogue entries: 147
  missing key    -> KeyError (rejected)
  null           -> AttributeError (rejected)
  empty string   -> ValueError (rejected)
grills with an empty list in either unit: 0
```

That is stronger than what #385 established. I had only shown that no *current* entry produces an empty list, which left "a future vendor drop could" open — and with `min()` on an empty sequence raising rather than degrading, that would have been the argument against option 1. `from_dict` raising at catalogue load closes it: the bad data cannot reach an entity property, and nothing in this repo builds a `Grill` any other way.

**The tests bite.** Three mutations, all caught:

```
min_temp -> max of the list              caught (7)
max_temp -> min of the list              caught (7)
bounds ignore the unit (always F list)   caught (1)
```

That last one matters more than it looks — it is what stops the bounds silently reverting to the Fahrenheit list on a Celsius grill now that there is no conversion left in the file.

`climate.py` goes 89% → 97% branch coverage; the module total 95% → 96%. 180 passed on macOS natively, ruff/format/mypy clean, and all five CI checks green.

## One observation, not a blocker

`extra_state_attributes` keeps its `if accepted := ...` guard, and it is now the only uncovered line in `climate.py` (84). The snap at 168 keeps its guard too. Same unreachable condition, three sites, two of which still branch on it.

I would leave them: the guard there answers "should this attribute be published at all", which is a real question even if the empty case never arrives, whereas `min_temp` had to invent a number. Worth knowing that the dead branch simply moved rather than went away — if it ever reads as an inconsistency, 84 is the line to point at.

Merging.

PR #389, opened 2026-08-05
Name the grill in the discovery row

trailro:
`async_step_bluetooth` sets a title placeholder for the discovered grill:

```py
self.context["title_placeholders"] = {"name": self._device_id}
```

but nothing consumes it — `en.json` has no `flow_title`, so the frontend falls back to the integration's own title. Two grills waiting to be configured are two rows both reading "PitBoss", with nothing to tell them apart. This integration discovers by Bluetooth, so more than one pending at once is the ordinary case for anyone with a second grill.

One line in `en.json`. The test asserts the placeholder and the template that renders it *together*, because either one alone silently produces the same fallback — the pair is the invariant, not each half.

197 passed, ruff / ruff format / mypy clean. Verified the test fails without the translation key (KeyError), not just passes with it.

---

## A question rather than a change: `iot_class`

The manifest declares `local_push`. That was exactly right when Bluetooth was the only transport; with three it describes one of them:

| protocol | local or cloud | push or poll |
|---|---|---|
| `ble` | local | push — `subscribe_state` over notifications |
| `wss` (**`DEFAULT_PROTOCOL`**) | **cloud** | push — `subscribe_state` over the socket |
| `local` | local | **poll** — no push on the local endpoint, which is what motivated adaptive polling |

No single value covers all three, and the field takes one. Worth noting the default is `wss`, so the declared class does not describe the path a user gets unless they change it.

I have not touched it here — it is repo-level metadata, it shows in the HACS listing, and picking between `local_push` and `cloud_push` is a judgment about which users the manifest should describe rather than a bug with a right answer. Happy to send whichever you prefer as a one-line follow-up, or to leave it alone if the Bluetooth-first reading is the intended one.

### DIFF
--- a/custom_components/pitboss/translations/en.json
+++ b/custom_components/pitboss/translations/en.json
@@ -1,6 +1,7 @@
 {
   "title": "PitBoss",
   "config": {
+    "flow_title": "{name}",
     "abort": {
       "already_configured": "Grill is already configured",
       "reconfigure_successful": "Re-configuration was successful",
--- a/tests/test_config_flow.py
+++ b/tests/test_config_flow.py
@@ -1,3 +1,5 @@
+import json
+import pathlib
 import threading
 from unittest.mock import Mock, patch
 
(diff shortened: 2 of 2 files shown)


dknowles2:
Premise verified rather than assumed — `main`'s `en.json` has `config` keys `['abort', 'error', 'step']` and no `flow_title`, so the placeholder `async_step_bluetooth` sets really did go nowhere.

The fix is the documented convention, not an invention: **181 core integrations declare `config.flow_title`**, and `philips_js` uses the identical `"{name}"` template.

```
tesla_wall_connector  {serial_number} ({host})
deconz                {host}
philips_js            {name}
```

Confirmed the test bites, by deleting the key rather than trusting the claim:

```
without flow_title: CAUGHT | E  KeyError: 'flow_title'
```

Asserting the pair is the right call and worth keeping that way. `flow_title` is rendered by the frontend, not by any Python this suite can reach — it appears nowhere in `homeassistant.helpers.translation` or the config panel's Python — so a backend test can only pin the two halves against each other. Renaming the placeholder to `device` would leave `format(**placeholders)` raising, which is exactly the failure that matters.

197 passed, ruff / ruff format / mypy clean, and all CI checks green.

Merging.

## On `iot_class`

You are right that `local_push` no longer describes the default path, and right not to have changed it here.

For whoever picks it up: HA takes one value and there is no combination form, so every multi-transport integration is already approximate — `esphome` and `shelly` both declare `local_push`, `tuya` declares `cloud_push`. The field feeds the HACS listing and the integration docs, so the question is which reading is least misleading to someone browsing, not which is technically complete.

My read is that `local_push` should stay, for a reason worth writing down: `iot_class` describes what the integration *is capable of*, and this one can run fully local with push over Bluetooth. `wss` being `DEFAULT_PROTOCOL` is a setup default, not a capability ceiling — and now that `local` exists, "this can be local" is more true than it was, not less. Changing it to `cloud_push` would tell a browsing user the opposite of what the local WiFi transport just made possible.

Not my call either, so leaving it — @dknowles2.

PR #391, opened 2026-08-05
Create the error sensors a board actually reports

### LINKED ISSUE #388: "Startup error" binary sensor sits permanently unknown on PBM/PBM2 boards
## The gap

`binary_sensor.py` creates one entity per key in `ENTITY_DESCRIPTIONS` for every grill, but not every board reports every key. Sweeping the live code of all reachable boards' parsing routines against the 12 binary-sensor keys shows exactly one mismatch:

- **`erL` ("Startup error") is never emitted by PBM and PBM2** — neither routine assigns it, on either board generation. Every other key is universal.

On those boards the entity is created anyway and sits **permanently unknown**: the coordinator's merged state never carries `erL`, so `is_on` returns `None` forever. Affected models (per `grills.json` today): PBM — Lexington Wi-Fi Upgrade, PB1000NXW, PB1020CS2, PB1020NXW, PB1230CS1, PB1250CS, PB1450CS, PB1500NXW, PB1600CS, PB1600SPW, PB1600SPW2, PB820CS1, PB850AMB, PB850CS1, PB850CS2, PB850M; PBM2 — PB1500NXW, PB1600 CS, PB850CS2, PB850M.

Reproduced end-to-end with synthesized frames run through the boards' real parse routines: all other sensors populate, `binary_sensor.*_startup_error` stays unknown for the life of the entry.

## The fix, and its dependency

The board-knowledge boundary (REVIEW.md) says facts about boards belong in pytboss, not in a hardcoded board list here. **This is blocked on dknowles2/pytboss#599**, which adds exactly the primitive needed:

```py
coordinator.api.spec.control_board.emits("erL")  # False on PBM/PBM2
```

Once that lands in a pytboss release, the change here is small: in `binary_sensor.py`'s `async_setup_entry`, create each entity only when the board emits its key. Written against `emits()` generically, it also self-heals any future per-board gap a `grills.json` refresh introduces, with no board names in this repo.

Sequencing:
1. dknowles2/pytboss#599 merged
2. pytboss release cut (main currently carries 9+ unreleased commits, including the `ftoc(null)` and unconverted-fields fixes — one release unblocks all of it)
3. double pin bump here (`requirements.txt` + `manifest.json`, per AGENTS.md)
4. the gating PR — happy to send it once 1–3 are done

trailro:
Fixes #388.

PBM and PBM2 have no `erL` in either parsing routine, so their "Startup error" sensor was created and then sat unknown for the life of the entry — the merged state never carries the key. Twenty models across the two boards.

```py
board = coordinator.api.spec.control_board
...
if not board.emits(entity_description.key):
    continue
```

Asked of the board rather than gated on a list of board names here, so a definitions refresh that adds or drops a field is followed without a change in this repo. `emits()` is dknowles2/pytboss#599, released in `2026.8.5` — already pinned in both files by #390, so nothing to bump here.

`erL` is the only one of these twelve flags that is not universal today; the sweep behind #388 found no other mismatch, and `primeState` — the other non-universal key — backs the primer switch, which `switch.py` already gates on the command being declared.

## Notes

- **The module-level `model` parametrization had to go.** The file now needs a second board, and pytest rejects a per-test `parametrize` for an argument already parametrized at module level. Converted to per-test decorators, which is what `test_sensor.py`, `test_switch.py`, `test_light.py` and `test_number.py` already do — the module-level mark in this file was the outlier. No test changed meaning; every existing one still runs against `PBV4PS2`.
- The new test also asserts two sensors that *should* exist still do, so a gate that thins out the platform fails rather than passes quietly.

198 passed, ruff / ruff format / mypy clean. Verified the test fails with the gate removed, not just passes with it in.

### DIFF
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -120,7 +120,16 @@ async def async_setup_entry(
     coordinator: PitBossDataUpdateCoordinator = hass.data[DOMAIN][entry.entry_id]
     entities: list[BinarySensorEntity] = []
     assert entry.unique_id is not None
+    board = coordinator.api.spec.control_board
     for entity_description in ENTITY_DESCRIPTIONS:
+        # Not every board reports every flag: PBM and PBM2 have no `erL` in
+        # either parsing routine, so their "Startup error" sensor was created
+        # and then sat unknown for the life of the entry. Asked of the board
+        # rather than gated on a list of names here, so a definitions refresh
+        # that adds or drops a field is followed without a change in this
+        # repo.
+        if not board.emits(entity_description.key):
+            continue
         entities.append(BinarySensor(coordinator, entry.unique_id, entity_description))
     entities.append(ConnectivitySensor(coordinator, entry.unique_id))
     entities.append(MeatProbeControlSensor(coordinator, entry.unique_id))
--- a/custom_components/pitboss/binary_sensor.py
+++ b/custom_components/pitboss/binary_sensor.py
@@ -11,8 +11,9 @@ from homeassistant.components.binary_sensor import (
     BinarySensorEntityDescription,
 )
 from homeassistant.config_entries import ConfigEntry
-from homeassistant.const import EntityCategory
+from homeassistant.const import EntityCategory, Platform
 from homeassistant.core import HomeAssistant
+from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 
 from .const import DOMAIN, probe_label
@@ -121,6 +122,7 @@ async def async_setup_entry(
     entities: list[BinarySensorEntity] = []
     assert entry.unique_id is not None
     board = coordinator.api.spec.control_board
+    registry = er.async_get(hass)
     for entity_description in ENTITY_DESCRIPTIONS:
         # Not every board reports every flag: PBM and PBM2 have no `erL` in
         # either parsing routine, so their "Startup error" sensor was created
@@ -129,6 +131,18 @@ async def async_setup_entry(
         # that adds or drops a field is followed without a change in this
         # repo.
         if not board.emits(entity_description.key):
+            # The affected grills are the ones already running, so the row is
+            # already in the registry. Home Assistant keeps a row whose entity
+            # stops being created and renders it as unavailable, which would
+            # turn "permanently unknown" into "permanently unavailable" for
+            # exactly the users this is meant to help. Removed here so the
+            # entity goes away on the next start rather than having to be
+            # deleted by hand.
+            unique_id = f"{entity_description.key}_{entry.unique_id}"
+            if stale := registry.async_get_entity_id(
+                Platform.BINARY_SENSOR, DOMAIN, unique_id
+            ):
+                registry.async_remove(stale)
             continue
         entities.append(BinarySensor(coordinator, entry.unique_id, entity_description))
     entities.append(ConnectivitySensor(coordinator, entry.unique_id))
--- a/tests/test_binary_sensor.py
+++ b/tests/test_binary_sensor.py
@@ -5,6 +5,7 @@ from unittest.mock import Mock
 import pytest
 from conftest import get_entity
 from homeassistant.core import HomeAssistant
+from homeassistant.helpers import entity_registry as er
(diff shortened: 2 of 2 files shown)


dknowles2:
The gate is right and the scope is exactly what you claim. Verified against every board rather than the two named:

```
boards: 17   models: 137

boards losing an entity, and which:
  PBM   ['erL']  -> 13 models
  PBM2  ['erL']  -> 4 models

entities removed in total: 17
boards unaffected: 15
```

So it removes precisely one flag on two boards and thins out nothing else — which is what the "the rest still arrive" half of the new test is guarding, and worth having.

The `pytestmark` conversion is right too. It is a real constraint, not a preference: pytest refuses a per-test `parametrize` for an argument already parametrized at module level, and this file was the outlier against `test_sensor.py`, `test_switch.py`, `test_light.py` and `test_number.py`.

198 passed, ruff / ruff format / mypy clean locally.

## One thing to add before this goes in: the existing installs

The 17 affected models are, by definition, the ones already running. Their `startup_error` entity is in the registry, and nothing here removes it. Tested by seeding the row a previous release would have left and setting the entry up:

```
seeded registry entry: binary_sensor.mygrill_startup_error
still in registry after setup: True
state after setup: None
```

A registry row with no backing entity does not disappear — HA keeps it and renders it as unavailable/restored, with manual deletion the only way out. So for the users who filed #388, "sits permanently unknown" becomes "sits permanently unavailable". Better, in that it is at least flagged and deletable, but not the outcome the PR describes; that outcome is only reached on a **fresh** install.

It is a few lines, and I prototyped it against your branch to be sure it works rather than just suggesting it:

```py
if not board.emits(entity_description.key):
    unique_id = f"{entity_description.key}_{entry.unique_id}"
    if existing := registry.async_get_entity_id(
        "binary_sensor", DOMAIN, unique_id
    ):
        registry.async_remove(existing)
    continue
```

with `registry = er.async_get(hass)` alongside the `board` lookup. Same seeded-row test then gives:

```
still in registry after setup: False
```

and the rest of the suite is unaffected — 198 passed with it applied. I have reverted it; it is yours to write, and worth a test of its own since the seeded row is the only way this path gets exercised.

Not merging until that lands, since #388's complaint is about what existing users see and this is the half that reaches them. Everything else here is ready.

trailro:
Good catch, and it is the half that reaches the people who filed it — thanks for prototyping it rather than describing it. Landed at `6750973`.

Confirmed the mechanism myself before writing it: the stale row survives setup and the entity renders unavailable, so the outcome the PR described was reachable only on a fresh install.

Two small differences from your sketch:

- `Platform.BINARY_SENSOR` rather than importing `DOMAIN as BINARY_SENSOR_DOMAIN` — with the alias, `ruff check --fix` splits the `homeassistant.components.binary_sensor` import into two separate blocks to sort it. Same value, and `Platform` is already the idiom for this argument.
- The test seeds the row through `mock_config_entry` before setup, since that is the only way this path runs, and also asserts a flag the board *does* report keeps its row — so a removal loop that reached too far fails rather than passes quietly.

Verified it fails with the removal block taken out, not just passes with it in. 199 passed, ruff / ruff format / mypy clean.

dknowles2:
`6750973` closes it, and the two changes you made to the sketch are both improvements on it.

**`Platform.BINARY_SENSOR` over the alias.** Checked the reason rather than taking it — with `DOMAIN as BINARY_SENSOR_DOMAIN`, `ruff check --fix` really does split the import:

```
from homeassistant.components.binary_sensor import DOMAIN as BINARY_SENSOR_DOMAIN
from homeassistant.components.binary_sensor import (
```

Two blocks from the same module, which is worse than what it replaced. `Platform` is the right call and would have been even without that.

**Asserting a reported flag keeps its row** is the better half of your test. A removal loop that reached too far is the realistic way this goes wrong, and my version would not have caught it. Confirmed it does:

```
removal block deleted                          caught
gate inverted (keep only what is NOT emitted)  caught
removal reaches too far: drop every row        caught   <- your assertion
gate always passes                             caught
```

199 passed, ruff / ruff format / mypy clean locally, and all five CI checks green.

For the record on what this ships: 17 models across PBM and PBM2 lose their Startup error sensor — the row goes on the next start rather than needing manual deletion — and 15 boards are untouched. `erL` is the only non-universal flag of the twelve, so nothing else moves.

Merging.

PR #393, opened 2026-08-06
Convert the held probe targets when the grill's unit flips

trailro:
Found sweeping the released stack (pytboss `2026.8.5` + `2026.8.1`) with synthesized frames through the boards' own parsing routines.

## The window

`probe_targets` holds values in the grill's unit, and the poll is the only path that refreshes it — `_on_state_update` applies the poll interval and publishes the state, but never touches the targets. A unit change arriving by push therefore leaves the old-unit numbers in place for up to a poll interval (10 s lit, 60 s in standby). In that window:

- a probe target number shows the old value labeled with the new unit — **74 °F** where 74 °C was set;
- target-reached compares across units: a probe reading **140 °F** sits above a target of **74** held from Celsius, so the sensor fires for a probe that is 14 °C short of done.

Only probes whose target is not board-reported are exposed — `probe_target()` prefers the state's own `pNTarget`, which arrives already flipped — but that is most probes on most boards: `p2Target` is emitted by 26 of 137 models, `p3`/`p4` by none.

## The fix

Converted rather than dropped, so the target stays on the dashboard; the next poll re-reads what the grill is actually holding. Done in `async_update_listeners`, where the pending setpoint already expires for the same reason — the state these decisions depend on arrives by three paths, and all three end there. `restored_targets` needs nothing: it is kept in Fahrenheit for exactly this reason, per its own comment.

The unit is tracked from `None` rather than seeded from `grill_unit` at construction, since the unit read before any data is the Fahrenheit default, not something the grill said.

The test drives the flip through `async_set_updated_data` (the push path), asserts 74 °C → 165 °F, and flips back — a conversion that drifted (floor instead of round) would fail the round trip.

211 passed, ruff / ruff format / mypy clean. Verified the test fails without the conversion, not just passes with it.

### DIFF
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -138,7 +138,12 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         Celsius meant presenting one unit while commands were snapped in the
         other.
         """
-        if (data := self.data) and not data.get("isFahrenheit", True):
+        return self._unit_of(self.data or StateDict())
+
+    @staticmethod
+    def _unit_of(state: StateDict) -> str:
+        """The unit a state frame is expressed in."""
+        if not state.get("isFahrenheit", True):
             return UnitOfTemperature.CELSIUS
         return UnitOfTemperature.FAHRENHEIT
 
@@ -348,6 +353,14 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         try:
             # Copied: we add to this below, and it is not ours to mutate.
             self.probe_targets = dict(await self.api.get_probe_targets())
+            # Recorded at the point of the read rather than left for
+            # `_follow_probe_targets_unit` to infer: these values came back
+            # in the unit of the frame that carried them, and `self.data` is
+            # not updated until this poll returns -- so the inference would
+            # see a flip that has already been accounted for and convert
+            # them a second time. `state` is the merged frame, so a partial
+            # one that omits the flag still carries the previous value.
+            self._targets_unit = self._unit_of(state)
         except Exception as ex:  # noqa: BLE001
             self.logger.debug("Could not fetch the probe targets: %s", ex)
             return
--- a/custom_components/pitboss/coordinator.py
+++ b/custom_components/pitboss/coordinator.py
@@ -86,6 +86,11 @@ class PitBossDataUpdateCoordinator(DataUpdateCoordinator[StateDict]):
         self.probe_targets: dict[int, int] = {}
         self.restored_targets: dict[int, int] = {}
         self._targets_seeded = False
+        # The unit `probe_targets` was last known to be in, so a unit change
+        # can convert the held values instead of serving them mislabeled.
+        # `None` until the first update: the unit read before any data is
+        # the Fahrenheit default, not something the grill said.
+        self._targets_unit: str | None = None
         # The grill setpoint we asked for, held until the grill confirms it.
         # Kept here rather than on the climate entity because the setpoint has
         # more than one reader -- the climate entity and the grill setpoint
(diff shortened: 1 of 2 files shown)


custom_components/pitboss/coordinator.py
```diff
@@ -213,15 +218,48 @@ def _expire_pending_setpoint(self) -> None:
...
+        `probe_target` prefers the state's own `pNTarget`, which arrives
+        already flipped -- but those are most probes on most boards.
+        `restored_targets` needs nothing: it is kept in Fahrenheit for
+        exactly this reason.
+        """
+        unit = self.grill_unit
+        if self._targets_unit is None or not self.probe_targets:
+            self._targets_unit = unit
+            return
+        prev, self._targets_unit = self._targets_unit, unit
```
dknowles2, line 250: This infers "the unit changed, so what I hold is stale" — but on the poll path `_async_refresh_probe_targets` has already replaced `probe_targets` with values in the *new* unit before this runs, so the inference is wrong exactly when a flip arrives by poll. `_targets_unit` is never updated there, because `grill_unit` still reads the old `self.data` at that point.

Recording the unit at the point of the read removes the guesswork:

```py
self.probe_targets = dict(await self.api.get_probe_targets())
self._targets_unit = self._unit_of(state)
```

with `_unit_of` factored out of `grill_unit` so both read a frame the same way.

dknowles2 (changes requested):
Verified the bug you describe is real, and the fix is right for the path it covers — I drove a unit flip through `async_set_updated_data` and confirmed the held targets come out converted, and that your test fails without the conversion rather than merely passing with it. The reasoning about `restored_targets` needing nothing is correct, and tracking the unit from `None` rather than seeding it from `grill_unit` is the right call for the reason you give.

One problem, on the other path. `_async_refresh_probe_targets` replaces `probe_targets` wholesale on **every** poll while the grill is lit, with values the grill reports in its *current* unit — and it runs before `self.data` is updated, so `grill_unit` inside it is still the old unit. When a flip arrives by poll, the targets are re-read already correct, and then `_follow_probe_targets_unit` converts them a second time.

A test on your branch, driving a real refresh rather than the push path:

```py
async def test_a_unit_flip_arriving_by_poll(
    coordinator: PitBossDataUpdateCoordinator, mock_pitboss: Mock
) -> None:
    """The poll re-reads the targets from the grill, already in the new unit."""
    coordinator._api_started = True
    mock_pitboss.is_connected.return_value = True

    mock_pitboss.get_state.return_value = StateDict(moduleIsOn=True, isFahrenheit=True)
    mock_pitboss.get_probe_targets.return_value = {3: 165}
    await coordinator.async_refresh()
    assert coordinator.probe_target(3) == 165

    # Panel flipped to Celsius. The next poll carries the new flag, and
    # `get_probe_targets` answers in the grill's unit -- 74 C, not 165 F.
    mock_pitboss.get_state.return_value = StateDict(moduleIsOn=True, isFahrenheit=False)
    mock_pitboss.get_probe_targets.return_value = {3: 74}
    await coordinator.async_refresh()

    assert coordinator.probe_target(3) == 74
```

```
E       assert 23 == 74
```

It passes on `main` and fails on this branch, so it is a regression rather than something already there. The symptom is the one you set out to fix, moved to the more common path: a target number showing 23 where 74 was set, and target-reached comparing a probe against it, for one poll interval.

A fix that works — full suite green at 211, including your own test:

```diff
@@ class PitBossDataUpdateCoordinator
-        if (data := self.data) and not data.get("isFahrenheit", True):
+        return self._unit_of(self.data or StateDict())
+
+    @staticmethod
+    def _unit_of(state: StateDict) -> str:
+        """The unit a state frame is expressed in."""
+        if not state.get("isFahrenheit", True):
             return UnitOfTemperature.CELSIUS
         return UnitOfTemperature.FAHRENHEIT
@@ async def _async_refresh_probe_targets
             self.probe_targets = dict(await self.api.get_probe_targets())
+            self._targets_unit = self._unit_of(state)
```

The targets just read are in the unit of the frame they came with, so record that rather than leaving `_follow_probe_targets_unit` to infer a change that has already been accounted for. `state` there is the *merged* frame — `_async_poll` calls `_merge_state` before passing it — so a partial frame that omits `isFahrenheit` still carries the previous value rather than falling back to the Fahrenheit default.

Worth a test on the poll path either way; the push path and the poll path refresh `probe_targets` by different routes, and only one of them was covered.

trailro:
You are right, and it is a regression I introduced rather than a gap I left: the branch fixed the push path and broke the poll one, which is the path that runs far more often. Reproduced your test on this branch before changing anything — `assert 23 == 74`, exactly as you had it — and confirmed it passes on `main`.

The mechanism is as you describe. `_async_refresh_probe_targets` replaces `probe_targets` wholesale with values already in the new unit, and it runs *before* `self.data` is updated, so `grill_unit` inside it still reads the old frame and `_targets_unit` is never moved. `_follow_probe_targets_unit` then sees a flip that has already been accounted for and converts a second time.

Taken as proposed, at `6f6fe23`: `_unit_of` factored out of `grill_unit`, and the unit recorded at the point of the read rather than inferred afterwards. Recording it is the better shape for the reason your diff implies — the values and their unit arrive together, so nothing has to be reconstructed from state that has since moved on.

Your note that `state` is the merged frame is what makes it safe, and I checked it rather than taking it on trust: `_async_poll` calls `_merge_state` before passing it here, so a temperatures-less frame carries the previous `isFahrenheit` instead of falling back to the Fahrenheit default — which would otherwise convert Celsius targets on every partial poll.

Your test is in, attributed in the commit message. It fails without the change and passes with it, and it covers the path my own test did not: the push path and the poll path reach `probe_targets` by different routes, and I had only pinned one.

212 passed, `mypy .` clean over the whole tree, `ruff` / `ruff format` clean.

dknowles2 (approved):
Took the fix as proposed, and checked the things worth checking rather than taking the summary on trust.

Verified on `6f6fe23`: 212 pass, `mypy .` clean over the whole tree, `ruff`/`ruff format` clean. Removing the `_targets_unit = self._unit_of(state)` line fails exactly one test, so the poll-path coverage is load-bearing rather than decorative.

Also merged this against #394 locally before landing either — both touch `_async_refresh_probe_targets`, and the combination is what neither branch can test on its own. The source merges clean and the combined suite is 218 green. The floor conversion in #394 and the unit conversion here agree by construction, since `probe_target_floor` tracks the same unit `_follow_probe_targets_unit` converts into: 50 F and 10 C map to each other exactly.

Good catch on confirming `_merge_state` runs first rather than taking my word for it — that is the part that would have bitten silently, on partial frames only.

PR #394, opened 2026-08-06
Treat a probe target at the minimum as no target

trailro:
Found on a live grill, then narrowed to one variable on the same grill.

## What happens

The board reports its floor — 50 °F, `DEFAULT_PROBE_MIN_TEMP` — for a probe it is not working to. A probe sitting in a room is warmer than that, so `target_reached` fires the moment a probe is plugged in:

```
08:53:58  sensor.…_probe_1      unknown → 25 °C     (probe plugged in)
08:53:58  …_mpc_target_reached  off     → on        (same second)
          number.…_probe_1_target  10 °C           (unchanged, = 50 °F)
          switch.…_module_power    off             (cold, switched off)
```

The sensor was reporting the probe's *presence*, not the meat's temperature.

## Which variable it actually is

Two candidates: the probe being absent, or the grill being off. Both were ruled out on the same grill within minutes — with the probe plugged in and the grill still off, a real target sticks:

```
08:56:28  set 63 °C
08:56:35  number → 62 °C   ·  mpc_target_reached → off   (probe 26 °C < 62 °C)
08:58:35  still 62 °C, across several polls
```

(63 → 62 is the board's own round trip: 63 °C is 145 °F, and back is `floor((145-32)/1.8)` = 62. Not this repo's arithmetic — probe targets do not go through a snap.)

So the floor is what the board says when it has no target, and the real value appears as soon as it has one.

## Why dropping it is safe

At the floor the answer carries nothing either way. As the board's placeholder, target-reached is true for every probe in a room. Asked for deliberately, it was already true before it was asked — no probe warm enough to cook was ever below 50 °F. There is no reading of "target = 50 °F" under which firing the sensor informs anyone.

## Where

In `probe_target()` — the one reader both the number and the binary sensor go through — rather than in the sensor, so the number stops offering the placeholder as a setting too. A restored floor is no longer seeded to the grill for the same reason: it would write the board's own placeholder back to it.

It applies to every probe, by three routes: probe 1 through the board's own `p1Target` on every model, probe 2 the same way on 26 of them, and any probe through a held or restored value.

215 passed, ruff / ruff format / mypy clean. Verified the four new tests fail without the change, not just pass with it.


dknowles2 (commented):
The evidence here is good and the narrowing is exactly right — ruling out "probe absent" and "grill off" on the same grill within minutes is what makes the conclusion stick, and the 63 → 62 aside correctly identifies that as the board's own round trip rather than anything in this repo. I checked: 215 pass, `mypy .` clean over the whole tree, `ruff`/`ruff format` clean. Putting it in `probe_target()` rather than the sensor is the right level, since the number and the binary sensor both come through there.

One consequence I do not think is intended.

`DEFAULT_PROBE_MIN_TEMP` is not only the board's placeholder — it is also the probe number's `native_min_value`. So the control offers the user exactly the value this now discards. Driving it through the entity:

```
slider minimum offered to the user: 50
right after setting 50: 50
after the settle window: unknown
```

The set is sent, the pending value holds it for `MCU_SETTLE_SECONDS`, then `probe_target()` returns `None` and the entity goes unknown. A slider whose own minimum silently fails to stick reads as a bug to whoever hits it, even though nothing downstream is wrong.

Your argument covers why *target-reached* should not fire at 50 F, and I agree with it — nothing warm enough to cook was ever below that. It does not cover the number continuing to advertise 50 as settable. The complement would be raising the control's floor so the placeholder is never on the dial:

```py
@property
def native_min_value(self) -> float:
    # One above the board's own "no target" placeholder, which
    # `probe_target` drops -- offering it would mean offering a value
    # that cannot stick.
    return TemperatureConverter.convert(
        DEFAULT_PROBE_MIN_TEMP + 1,
        UnitOfTemperature.FAHRENHEIT,
        self.native_unit_of_measurement,
    )
```

Or a separate constant, if you would rather the placeholder and the control's floor not be tied to each other at all — they are different facts that happen to share a number today, which is part of why this is easy to miss.

One question rather than a finding: "the board reports its floor when it has no target" is a fact about the hardware, true with no Home Assistant in the picture, so there is an argument it belongs in pytboss's `get_probe_targets` rather than here. I lean towards here — mapping "no target" onto entity state is presentation, and pytboss returning what the board said is defensible — but @dknowles2 may see it differently, and it would be worth settling before this pattern gets copied.

Not blocking on the pytboss question. The `native_min_value` one I would fix before merging.

custom_components/pitboss/coordinator.py
```diff
@@ -245,7 +246,23 @@ def probe_target(self, probe_number: int) -> int | None:
...
+        for probe 1 with nothing set, and the real value the moment one was
+        -- and at that value the answer carries nothing either way: as the
+        board's placeholder it makes target-reached fire for any probe
+        warmer than 50 F, which is every probe in a room; and asked for
+        deliberately it was already true before it was asked. Dropped here
+        rather than in the sensor, so the number stops offering it as a
+        setting too.
         """
+        target = self._resolve_probe_target(probe_number)
+        if target is None or target == self._probe_target_floor:
```
dknowles2: This value is also `TargetProbeTemperature.native_min_value`, so the number offers it as a setting and then drops it: the slider shows 50, the set is sent, the pending value holds for the settle window, and the entity then goes `unknown`. Worth raising the control's floor one step above the placeholder in the same change, so nothing on the dial is unsettable.

custom_components/pitboss/coordinator.py
```diff
@@ -326,6 +352,10 @@ async def _async_refresh_probe_targets(self, state: StateDict) -> None:
             # Converted at seeding time, into whatever unit the grill has
             # *now* -- which may not be the unit it had at capture.
             temp = self._from_fahrenheit(held)
+            if temp == self._probe_target_floor:
```
dknowles2: This one is unambiguously right — seeding the board's own placeholder back to it would be writing a target that means "no target".

dknowles2:
Settling the open question from my review: @dknowles2 confirms this stays in the integration rather than moving to pytboss. Mapping the board's placeholder onto "no target" is presentation, and pytboss returning what the board actually said is the right behaviour for a library.

So the only outstanding item here is `native_min_value` still offering the value `probe_target` now drops.

trailro:
Fixed at `cdd1aac`, and you are right that it was unintended — I changed what `probe_target` returns without looking at what the control was still offering.

Taken slightly differently from your sketch, for the reason your second option raises. Rather than converting `DEFAULT_PROBE_MIN_TEMP + 1` from Fahrenheit, the dial now reads the coordinator's own floor:

```py
return float(self.coordinator.probe_target_floor + 1)
```

`TargetProbeTemperature.native_unit_of_measurement` is `coordinator.grill_unit`, which is the unit `probe_target` and the floor are already in — so nothing is converted here, and the dial's bottom and the value that gets discarded cannot drift apart by construction. The alternative, a separate constant, would have let them: they are different facts, but the second is *defined by* the first — the control's floor exists precisely because the placeholder is indistinguishable from "no target". Deriving it says so; a second literal would leave a reader to notice the coupling.

The placeholder property is public now (`probe_target_floor`) since the number is a legitimate second reader, with a docstring saying why.

`tests/test_number.py`'s existing assertion moves from `min == 50` to `51`. That is the intended change rather than a broken test, so I left a comment there pointing at the new one, which asserts the property rather than the number: the minimum offered must be above what `probe_target` discards, and the minimum that *is* offered survives being set.

On the pytboss question — agreed, and thanks for settling it. `get_probe_targets` returning what the board said is the right contract for a library; "no target" is a presentation decision, and it belongs where the entities are.

216 passed, `mypy .` clean over the whole tree, `ruff` / `ruff format` clean.

dknowles2:
Reviewed and ready on my side — this needs `main` merged in and then I will land it.

Deriving the dial's floor from `probe_target_floor + 1` rather than re-converting the constant is better than what I sketched, and for the reason you give: the dial and the discarded value are in the same unit by construction, so they cannot drift apart. Making the property public with a docstring that says why the number is a legitimate second reader is the right call too.

Verified rather than taken on trust: 216 pass on `cdd1aac`, `mypy .` clean over the whole tree, `ruff`/`ruff format` clean. Reverting `native_min_value` to `probe_target_floor` (no `+ 1`) fails two tests, so the new coverage bites.

## The merge

#393 landed first, and it appended a test to the end of `tests/test_coordinator.py` where this branch does too. **Source files merge clean** — only the test file conflicts, and it is purely additive: keep both blocks, in order, nothing on either side changes.

I resolved it locally to check the combination before landing either of them, since you and I could each only test one side:

- combined suite: **218 pass**, `mypy .` clean
- the two changes agree by construction — `probe_target_floor` tracks the grill's current unit, which is the same unit `_follow_probe_targets_unit` converts into, so 50 °F and 10 °C map to each other exactly and a converted held target lands on the floor in the new unit rather than one degree off it

I tried to push that resolution here to save you the round trip, since `maintainerCanModify` is on, but pushing to a fork is outside what I am allowed to do — so it is yours. It is a `git merge main` and deleting six conflict markers.

## Unrelated, worth knowing

Your parenthetical about 63 → 62 being the board's own round trip prompted me to settle it properly: dknowles2/pytboss#600. Measured at 46 of the 116 Celsius targets between 5 and 120, always low by exactly one, and not fixable at either end — the set command's payload is byte-identical whichever unit the grill is in, so the firmware converts rather than us, and the read side is the vendor's own routine run verbatim. Your reading was right; it is now written down so it stops being rediscovered.

dknowles2 (approved):
Re-verified on `8c6d89b` against current `main`, which has moved a fair way since my review — #395 through #399 all landed, including the entity-registry and real-state test work that touches the same platforms.

Rather than trust the CLEAN status: 407 pass, `mypy .` clean over the whole tree, `ruff`/`ruff format` clean. Both of your gates still bite after four merges of main — reverting `native_min_value` to `probe_target_floor` fails 2, and dropping the floor check out of `probe_target` fails 4.

Also checked the interaction with #399's new per-board tests, since those drive the entities from real parsed state across nine control board families and could plausibly have collided with a change to what `probe_target` returns. They don't: `build_state` seeds `p1Target` at 191 rather than the placeholder, and the invariant test exempts probe targets for exactly the reason your change makes sharper — a target nobody has set is legitimately absent.

Deriving the dial's floor from `probe_target_floor + 1` remains better than the constant conversion I suggested. Thanks for the patience through four rebases; the merges were my doing, not yours.

